PR #62, opened 2019-05-20
fix: potential panic if send returns nil resp and no error

shuoranliu:
It is unacceptable for sendToMetaPatition to return nil resp packet and
no error, which may lead to callers to panic.


shuoranliu:
typo fixed. "faild" -> "failed"

PR #311, opened 2019-12-15
fix: broken directory mtime

shuoranliu:

The mtime of parent inode is not updated when create or delete a dentry.

This commit fixes the issue by updating mtime of parent inode in server
side and evict inode cache in client side.

**Release note**:


awzhgw:
@shuoranliu  please plus your signature information on commit

PR #316, opened 2019-12-18
Add CNCF reference to README

caniszczyk:
This is required per CNCF policies: https://github.com/cncf/foundation/blob/master/website-guidelines.md

Signed-off-by: Chris Aniszczyk <caniszczyk@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 
 ## Overview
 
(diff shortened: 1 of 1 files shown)


caniszczyk:
Actually I noticed the README was already updated, but it's nice to include this information at the top of the README

Issue #393, opened 2020-02-28
Feature request: ARM and ARM64 architecture support


dereulenspiegel:
I had a quick look at this project and I like it. However I saw only x86_64 images on docker hub. Wouldn't it be easily possible to build this project for ARM/ARM64 and provide multi arch images?
This would probably be awesome for everyone who tries to run cluster with k3s and similar offerings on ARM clusters.

awzhgw:
@dereulenspiegel  Thank you for your suggestion
We Will try to do this thins.

mervinkid:
Hi @dereulenspiegel ,

Thanks for your feedback.

Unfortunately, we have not adapted ChubaoFS to the ARM and ARM64 platform yet.

We hope that ChubaoFS can help users in more scenarios. But we are not very familiar with the scenarios of these platforms you mentioned. 

Can you elaborate on the dozens of user scenarios and requirements of these platforms you mentioned.

dereulenspiegel:
Hi @mervinkid 

thank you for your quick response. There are multiple scenarios where a distributed storage system on ARM(64) would be valuable.
First there are more and more cloud provider offering ARM based servers (Amazon, Scaleway, probably a few others). Making ChubaoFS available for ARM based platforms would make it the prime choice for running distributed storage on pure ARM or even mixed ARM/x86 clusters.
Also the need for edge computing is pushing for more and more infrastructure on the like. In these scenarios x86 is not always the best choice. Bringing K8s/k3s with reliable distributed storage would very likely benefit many scenarios on the infrastructure edge.
Lastly a small ARM based k8s cluster is great development environment. Unfortunately we are lacking reliable distributed storage here.

I hope that this helps assessing if a port to ARM based platforms would be worth it for you. I surely hope so.

mervinkid:
Hi @dereulenspiegel 
Thank you for your explanation, I understand and think this is interesting.
Do you have any experience in adapting the Go language project to ARM and ARM64 platforms?
We would love to have a deeper discussion with you.
If you like, we also hope you can join and complete this adaptation.

dereulenspiegel:
If ChubaoFS only has Go based dependencies, then porting to ARM/ARM64 should be quite easy. Go itself has cross compiling capabilities, simply by setting the environment variable `GOARCH` to `arm` or  `arm64` should be sufficient to tell the compiler to produce ARM-compatible binaries. Only if cgo is involved, things get more complicated.
[...]

PR #398, opened 2020-03-03
Replaced Dup2 syscall with Dup3

dereulenspiegel:
The Dup2 syscall doesn't seem to be supported on all architectures on Linux. The newer Dup3 syscall however seems to have a much wider architecture support. This PR replaces all occurences of Dup2 with Dup3 to help push #393 forward.
Unfortunately the file `vendor/github.com/tiglabs/raft/util/log/log_crash_unix.go` still contains a call to Dup2 which isn't behind a build flag. So either this vendored dependency needs to be patched locally in the vendor folder (bad), or this is fixed upstream or the dependency is replaced by something else.

### DIFF
--- a/client/fuse.go
+++ b/client/fuse.go
@@ -139,7 +139,7 @@ func main() {
 	}()
 	syslog.SetOutput(outputFile)
 
-	if err = syscall.Dup2(int(outputFile.Fd()), int(os.Stderr.Fd())); err != nil {
+	if err = syscall.Dup3(int(outputFile.Fd()), int(os.Stderr.Fd()), 0); err != nil {
 		daemonize.SignalOutcome(err)
 		os.Exit(1)
 	}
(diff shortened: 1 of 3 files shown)


shuoranliu:
@dereulenspiegel 

Thanks for the pull request, and we appreciate your efforts in promoting the projects to more architectures.

As I checked the "dup3" syscall, it seems to be Linux-specific so it won't compile on macOS I guess. Since we’ve gone this far, how about taking it a little bit further that we provide an arch and OS free function named "dup"? Please check storage/fallocate_*.go for your reference which avoids the compile issue on macOS due to "fallocate" syscall. And we are more than happy to hear your opinions on how to deal with such issues in other projects.

Look forward to hearing from you. Thanks in advance.

mervinkid:
@shuoranliu 

I wrapped the `syscall.Dup2` method. Now users can compile on MacOS.

PR #471, opened 2020-04-07
Enhancement: add support for 'if-match' precondition in header object…

### LINKED ISSUE #459: Feature request: conditional parameter support for S3-compatible interfaces

The Amazon AWS S3 service itself provides some conditional judgment parameters to reduce the complexity of user maintenance data.
The condition parameter allows the user to determine whether the data needs to be updated or the version of the data without reading the data.
A typical case is that users store a large amount of vocabulary data on ChubaoFS, which is maintained by a business service. The service for maintaining data needs to update the vocabulary data. In order to selectively update and avoid overwriting the data that does not need to be updated, the new version of the data hash must be told to the storage system through a simple interface call. Whether it is consistent with the new version.

ChubaoFS 'S3 compatible object storage interface needs to implement the same condition judgment parameters as Amazon AWS S3 service to improve compatibility.

Implement the following conditional judgment parameters in the **GetObject** and **HeadObject** interfaces:
| Parameter | Description |
| ---------- | ------------ |
| **If-Match** | Return the object only if its entity tag (ETag) is the same as the one specified, otherwise return a `412` (precondition failed).  |
| **If-Modified-Since** | Return the object only if it has been modified since the specified time, otherwise return a `304` (not modified). |
| **If-None-Match** | Return the object only if its entity tag (ETag) is different from the one specified, otherwise return a `304` (not modified). |
| **If-Unmodified-Since** | Return the object only if it has not been modified since the specified time, otherwise return a `412` (precondition failed). |

Reference:
- [API - GetObject - RequestSyntax](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html#API_GetObject_RequestSyntax)
- [API - HeadObject - RequestSyntax](https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html#API_HeadObject_RequestSyntax)

**Describe alternatives you've considered**
NONE.

**Additional context**
NONE.

LeeYubo:

Add support for condition headers when heading or getting an object.
Such as If-Match、If-Modified-Since、If-None-Match and If-Unmodified-Since.

fixes #459 

NONE.

**Release note**:
NONE.

### DIFF
--- a/objectnode/api_handler_object.go
+++ b/objectnode/api_handler_object.go
@@ -82,7 +82,7 @@ func (o *ObjectNode) getObjectHandler(w http.ResponseWriter, r *http.Request) {
 
 		var hyphenIndex = strings.Index(rangeOpt, "-")
 		if hyphenIndex < 0 {
-			_ = InvalidArgument.ServeResponse(w, r)
+			errorCode = InvalidArgument
 			return
 		}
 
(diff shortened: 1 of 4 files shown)


awzhgw:
@LeeYubo Conflicting files
objectnode/result_error.go  ,please resolve conflicts

PR #483, opened 2020-04-10
fixed: the test case for vol which does not use a exist zone

zhuhyc:
Signed-off-by: zhuhyc <zzhniy.163.niy@163.com>

N/A
N/A
**Release note**:
N/A

### DIFF
--- a/datanode/server.go
+++ b/datanode/server.go
@@ -51,7 +51,7 @@ var (
 )
 
 const (
-	DefaultZoneName         = "cfs_zone1"
+	DefaultZoneName         = proto.DefaultZoneName
 	DefaultRaftDir          = "raft"
 	DefaultRaftLogsToRetain = 10 // Count of raft logs per data partition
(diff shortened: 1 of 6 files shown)


awzhgw:
> There is not much problem with code logic, but I have disputes with business logic.
> 
> The purpose of specifying a zone is to allow those businesses that are sensitive to the zone where the data is located to specify their own optimal storage location, and users should not care about which zones are available in the cluster.
> 
> If no zone is specified, it means that the user has no requirements for the location of the data. I think one of the following two logics should be used:
> 
> 1. If no zone is specified, cross zone is performed by default.
> 2. If no zone is specified, select and use a zone with resources from the cluster, and balance the zone resources.


I think:
If no zone is specified, all dataPartition and metaPartition are allocated in a zone by default

If you specify a zone, you can let the user choose the zone to cross, and we only cross 2 zones

PR #500, opened 2020-04-16
Fix ObjectNode: When overwriting a file, if updating the new inode to…

awzhgw:
Fix ObjectNode: When overwriting a file, if updating the new inode to dentry is successful, the error results of the old inode unlink and

evict are ignored

### DIFF
--- a/objectnode/fs_volume.go
+++ b/objectnode/fs_volume.go
@@ -571,11 +571,13 @@ func (v *Volume) DeletePath(path string) (err error) {
 
 	// Evict inode
 	if err = v.ec.EvictStream(ino); err != nil {
-		return
+		log.LogWarnf("DeletePath EvictStream: path(%v) inode(%v)", path, ino)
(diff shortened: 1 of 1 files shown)


objectnode/fs_volume.go
```diff
@@ -700,7 +702,6 @@ func (v *Volume) AbortMultipart(path string, multipartID string) (err error) {
 		if _, err = v.mw.InodeUnlink_ll(part.Inode); err != nil {
 			log.LogErrorf("AbortMultipart: meta inode unlink fail: multipartID(%v) partID(%v) inode(%v) err(%v)",
 				multipartID, part.ID, part.Inode, err)
-			return
 		}
 		if err = v.mw.Evict(part.Inode); err != nil {
```
mervinkid, line 706: @awzhgw 
Discuss whether it is necessary to ignore evict errors.

PR #618, opened 2020-06-05
Enhancement: optimize head memory when query inode

mervinkid:

- Enhancement: optimize head memory when query inode.

### DIFF
--- a/metanode/partition_fsmop_inode.go
+++ b/metanode/partition_fsmop_inode.go
@@ -59,10 +59,10 @@ func (mp *metaPartition) fsmCreateLinkInode(ino *Inode) (resp *InodeResponse) {
 	return
 }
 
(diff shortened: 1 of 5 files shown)


metanode/partition_fsmop_inode.go
```diff
@@ -100,11 +100,6 @@ func (mp *metaPartition) getInodeTree() *BTree {
 	return mp.inodeTree.GetTree()
 }
 
-// Ascend is the wrapper of inodeTree.Ascend
-func (mp *metaPartition) Ascend(f func(i BtreeItem) bool) {
-	mp.inodeTree.Ascend(f)
-}
-
```
mervinkid, line 107: Removed unused function `metaPartition.Ascend(f func(i BtreeItem) bool)`.

PR #635, opened 2020-06-09
Enhancment: mark inode created by s3 api.

mervinkid:
Mark flag of inode to `0x2` if the inode created by S3 api.


awzhgw:
@mervinkid  please add unit test

PR #781, opened 2020-07-22
Refactor: optimize delete Extent on metanode and datanode

awzhgw:
Optimize delete Extent on metanode and datanode

- metanode: use batchDeleteExtent packet 
- datanode: batchDeleteExtent packet response time set to 120s
- datanode : add autoRepairLimiter Limit the speed of recovery

curl "masterAddr/admin/setNodeInfo?autoRepairRate=20"


mervinkid:
ACK and review in progress.

PR #810, opened 2020-08-04
feature: add near read support

yinlei-jinan:
Signed-off-by: yinlei-jinan <297155992@qq.com>

**Release note**:

```release-note
```

### DIFF
--- a/sdk/data/stream/stream_conn.go
+++ b/sdk/data/stream/stream_conn.go
@@ -56,6 +56,13 @@ func NewStreamConn(dp *wrapper.DataPartition, follower bool) *StreamConn {
 		}
 	}
 
+	if dp.ClientWrapper.NearRead() {
+		return &StreamConn{
+			dp:       dp,
+			currAddr: getNearestHost(dp),
+		}
+	}
+
 	epoch := atomic.AddUint64(&dp.Epoch, 1)
 	hosts := sortByStatus(dp, false)
 	choice := len(hosts)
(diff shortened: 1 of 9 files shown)


sdk/data/wrapper/wrapper.go
```diff
@@ -280,3 +286,33 @@ func (w *Wrapper) updateDataNodeStatus() (err error) {
...
+	return w.nearRead
+}
+
+// Sort hosts by distance form local
+func (w *Wrapper) sortHostsByDistance(hosts []string) []string {
+	var tmpHost string
+	for i := 0; i < len(hosts); i++ {
+		for j := i + 1; j < len(hosts); j++ {
+			if distanceFromLocal(hosts[i]) > distanceFromLocal(hosts[j]) {
+				tmpHost = hosts[i]
```
awzhgw, line 305: use host[i],host[j]=host[j],host[i] replace it

sdk/data/stream/stream_conn.go
```diff
@@ -185,3 +192,18 @@ func sortByStatus(dp *wrapper.DataPartition, selectAll bool) (hosts []string) {
 
 	return
 }
+
```
awzhgw, line 195: If the machine is turned on and down, the status is false, and after 1 minute, it becomes true again.. Here, getNearHost is called. If the first two copies are not available, will the machine be read eventually?

sdk/data/wrapper/wrapper.go
```diff
@@ -153,6 +156,9 @@ func (w *Wrapper) updateDataPartition(isInit bool) (err error) {
 	localLeaderPartitionGroups := make([]*DataPartition, 0)
 	for _, partition := range dpv.DataPartitions {
 		dp := convert(partition)
+		if w.followerRead && w.nearRead {
```
awzhgw, line 159: It cannot be sorted so simply. When updatingDataParttion, because dp.Host must be in the same order in the master-slave replication agreement. Unchangeable

sdk/data/wrapper/wrapper.go
yinlei-jinan: resolved

sdk/data/stream/stream_conn.go
yinlei-jinan: yes,  the machine who are not available will be try read after normal status machine

sdk/data/wrapper/wrapper.go
yinlei-jinan: add NearHosts to save sorted hosts

PR #840, opened 2020-08-18
Feature: add vol create, delete, read, write rate limiter in server

shijiesun:
Signed-off-by: shijiesun <ssj280@gmail.com>

in this PR, you can set createRate, deleteRate, readRate, writeRate in both client and server side. cfs-client pull server rate config periodly, then use min(clientRate, serverRate) as the real limitRate.
I add a admin api(/vol/rate) in master to set server rate, enhance the client api(/rate/set) to support createRate and deleteRate. 
I change the rate type from int64 to float64 to avoid the type cast forcely when setting limit using "rate" lib. 
createRate and deleteRate can limit the times of calling create and delete.
readRate and writeRate can limit the speed Rate(Byte/second) when calling read and write

check the ratelimit set in client
check the vol opertion in master server

**Release note**:
```
set createRate, deleteRate, readRate, writeRate config in master server
```


leonrayang:
Hi @shijiesun,

Thank you for this contribution on volume-level rate limiting (create/delete/read/write) with client-server min() merging and the `/vol/rate` admin API.

We are closing this PR because it has been stale for a long time and conflicts with current `master`. Since this PR was opened, CubeFS has evolved a more capable QoS system (`master/limiter.go`, client flow upload/assignment via `QosUpload`) for read/write flow and IOPS. The static four-rate model in this PR is largely superseded for data path limiting.

What is **not** yet covered in current master is **create/delete metadata OPS limiting** from this PR. If that is still needed, please open a focused issue/PR against current QoS or meta client limiter design rather than rebasing this entire change set.

Thanks again for your work!

PR #847, opened 2020-08-20
fix: batch inode get mechanism is out of service

shuoranliu:
Right now the batch inode get mechanism does not work since in when
getting attr, the inode cache is evicted every time the underlying extent
streamer does not exist. To eliminate the eviction in getAttr, we have
to trigger an active eviction when the underlying streamer quits
silently.

**Release note**:

```release-note
```


awzhgw (approved):
review+1

PR #849, opened 2020-08-21
Fix: dp snapshot function use current applyid   prevent raft painc

awzhgw:

Fix: dp snapshot function use current applyid   prevent raft painc


mervinkid (approved):
LGTM

PR #851, opened 2020-08-23
Enhance: datanode optmize performance  When there are many extents

awzhgw:

1. When using elatiseach to pressure test chubaofs, when writing and deleting continuously, a large number of extents will be generated under each dp, and these extents will be deleted continuously, the length of extentcache is only 40, when each extent is deleted , LoadExtentWithHeader will be loaded from the disk. When reading the header, it will read the 4096-byte header from the extentcrc file, thus causing some random reads. This pr is to optimize this phenomenon.

2. When there are many extents ，dataPartition computeUseSize  A lot of random IO was wasted, so The memory method is used to calculate the space used by dataPartition。

3. tinyExtentRepair donnot use limit by autoRepairLimit parameter

### DIFF
--- a/datanode/partition.go
+++ b/datanode/partition.go
@@ -512,21 +512,10 @@ func (dp *DataPartition) actualSize(path string, finfo os.FileInfo) (size int64)
 }
 
 func (dp *DataPartition) computeUsage() {
-	var (
-		used  int64
-		files []os.FileInfo
-		err   error
-	)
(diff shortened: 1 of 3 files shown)


mervinkid (approved):
Consider whether it is necessary to also maintain the real size of the tiny extent in memory.

mervinkid:
> the length of extentcache is only 40,
> How about adjust the value in config file?

Do you think the value of 40 is larger or smaller, and why?

PR #862, opened 2020-08-31
Fix: raft instance delete itself by applying ConfChange raft log.

### LINKED ISSUE #698: Refactor and enhance node offline process for DataNode and Metanode.
The current offline logic for DataNode and MetaNode relies on Raft algorithm and logic, which introduces the following problems:
1. When rejoining the partition that was removed, if Raft has a low number of logs, Raft will copy the logs and replay the logs to restore the partition data on the new node. The logs from which the partition was removed will be applied, causing problems with changes to the partition's replica members.
2. When the partition copy group has no leader, the partition copy members cannot be adjusted and the node cannot be taken offline.

It is necessary to adjust the offline logic of DataNode and MetaNode. When processing the logic of adjusting the members of the partition copy, it does not go through or rely on Raft related mechanisms. These adjustments may require proper reconstruction of the Raft module used by ChubaoFS.

ssscottt:
fixes #698


ssscottt:
do not merge

zhuhyc (approved):

PR #938, opened 2020-09-22
Fix: datapartition on delete normalextent action,forget remove it from

awzhgw:
Fix: datapartition on delete normalextent action,forget remove it from storage.extentInfoMap, so some dp can change to readonly status,now fix
this bug


mervinkid (approved):
LGTM

PR #944, opened 2020-09-23
Solved some  compatibility issues in multi zone feature

ssscottt:
1.use default value if parameters are empty in updating vol
2. set zone name of single zone volume as 'default' during loading metafile
3. fix bug in setting dp/mpRecoverPoolSize
4. add parameter crossZone to volume struct

Signed-off-by: xuxihao1 <xuxihao3@jd.com>


awzhgw (approved):
review+1

zhuhyc (approved):

PR #1011, opened 2020-11-09
Refactor: success when removing non-exist DP

mervinkid:

Refactor: success when removing non-exist DP.
> Response success to Master when removing a non-exists Data Partition.


awzhgw (approved):
review+1

PR #1018, opened 2020-11-12
add checksum for metanode snapshoting and add 2 config about it

### LINKED ISSUE #993: Bug report: metanode snapshot files lack of checksum verification
**Describe the bug**
When metanode starts and loads snapshot file, there is no checksum verification. Although snapshot file length is respected, a bit-rot problem still have the chance of being undetected, and therefore metanode would starts successfully with wrong meta data.

**Environment**
N/A

**To Reproduce**
Inject a bit-rot problem to the latest metanode snapshot file, and restart metanode.

**Expected behavior**
Checksum verification of the snapshot file which is very critical.

**Additional context**
Recommend to add checksum at the end of each snapshot file, and keep backward compatibility.

xrefft:
meatanode: add a checksum file for each snapshot file. forward compatible and support fallback

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #993 

Taking into account the forward compatibility of the snapshot file, we only create a new "checksum file" for each snapshot, instead of adding a checksum at the end of the snapshot file. For example, the resulting file-tree of snapshot dir will be :
```shell
.
├── .checksum.apply
├── .checksum.dentry
├── .checksum.extend
├── .checksum.inode
├── .checksum.multipart
├── .sign
├── apply
├── dentry
├── extend
├── inode
└── multipart
```
**Release note**:

```release-note
add configure "checksumFunc" for meta node：the value could be crc32|crc64|md5|sha1|sha256|sha512 (No case limit)
add configure "ignoreChecksumError" to neglect checksum not match error. (Once if the data is abnormal)
```

### DIFF
--- a/metanode/checksum.go
+++ b/metanode/checksum.go
@@ -1,0 +1,47 @@
+package metanode
+
+import (
+	"encoding/base64"
+	"fmt"
+	"github.com/chubaofs/chubaofs/util/errors"
+	"github.com/chubaofs/chubaofs/util/hashfactory"
+	"io/ioutil"
+	"strings"
+)
+
+func checksumToStr(sum []byte) string {
+	return base64.StdEncoding.EncodeToString(sum)
+}
+
+func strToChecksum(str string) ([]byte, error) {
+	return base64.StdEncoding.DecodeString(str)
+}
+
+func writeChecksumToFile(file string, sum []byte, typ string) error {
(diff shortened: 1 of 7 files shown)


xrefft:
We have tested and verified the following in the test environment as above:
- The correctness of cheksum logic.
- Forward compatibility of snapshot file format.
- Seamless fallback (only need to fallback binary).

PR #1072, opened 2021-01-06
fix: Authorize user with wrong policy but return success and update policy

xiangpingzhang:
Signed-off-by: zhangxiangping <240133996@qq.com>

**What this PR does / why we need it**when authorize a user with wrong policy,master should return error and  should not change the user's original policy

**Release note**:

```release-note
```

### DIFF
--- a/master/user.go
+++ b/master/user.go
@@ -271,7 +271,10 @@ func (u *User) updatePolicy(params *proto.UserPermUpdateParam) (userInfo *proto.
 		err = proto.ErrIsOwner
 		return
 	}
-	userInfo.Policy.AddAuthorizedVol(params.Volume, params.Policy)
+	if !userInfo.Policy.AddAuthorizedVol(params.Volume, params.Policy) {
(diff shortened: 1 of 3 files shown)


leonrayang:
Hi @xiangpingzhang,

Thank you for reporting and fixing the issue where authorizing a user with invalid policy strings could return success while still updating the user's policy.

We are closing this PR because it has been stale for a long time and conflicts with current `master`. However, the underlying concern appears still relevant: `AddAuthorizedVol()` silently drops unparseable entries and may persist an empty policy list without returning an error to the caller.

If you are still interested, please rebase onto current `master` and open a new PR that:
1. Validates policy entries **before** persisting (reject all-invalid input with a clear error).
2. Does not mutate the user's existing policy on validation failure.
3. Adds unit tests in `master/api_service_user_test.go` for invalid/empty policy cases.

We would be happy to review an updated fix.

Thanks again!

PR #1073, opened 2021-01-06
"Add:check duplicate policy when authorize user policy"

xiangpingzhang:
Signed-off-by: zhangxiangping <240133996@qq.com>

**What this PR does / why we need it**：check duplicate policy when authorize user policy

**Release note**:

```release-note
```

### DIFF
--- a/proto/user_proto.go
+++ b/proto/user_proto.go
@@ -198,19 +198,32 @@ func (policy *UserPolicy) RemoveOwnVol(volume string) {
 	}
 }
 
-func (policy *UserPolicy) AddAuthorizedVol(volume string, policies []string) { //todo check duplicate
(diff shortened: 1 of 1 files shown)


leonrayang:
Hi @xiangpingzhang ,

Thank you for this small but useful improvement to deduplicate policy entries in `AddAuthorizedVol()`.

We are closing this PR because it has been stale and is no longer mergeable as-is. The `// todo check duplicate` note in `proto/user_proto.go` is still present on current `master`, so the improvement has not landed yet.

If you'd like to contribute this again, a rebased PR against current `master` would be welcome — ideally with a unit test showing that duplicate permissions/actions in the request are stored only once.

Thanks again!

PR #1104, opened 2021-02-04
fix: build libcfs and java lib，upgrade arm dependent lib to lz4 1.9.3，zstd 1.4.8

hooklee2000:
fix build libsdk，upgrade arm dependent lib to lz4 1.9.3，zstd 1.4.8
Supporting   mvn  java lib to build/bin,change libsdk.so to libcfs.so


hooklee2000:
See #1108

PR #1115, opened 2021-02-26
fix: truncate file may cause garbage data for tiny extents

Victor1319:


mervinkid (approved):
Good.

mervinkid:
@shuoranliu 

I gave a simple review of the purpose and implementation logic of this pull request.
The code involved in the pull request needs to be tested before release.

PR #1166, opened 2021-06-21
fix: raft leader sending check quorum

shuoranliu:
This commit forces raft leader to send check quorum message to other
inactive nodes. Otherwise, when inactive nodes are back online, there is
a chance that they don't receive check quorum messages and therefore
their fsm remains "noleader" status which lead to unnecessary leader
election.

**Release note**:

```release-note
```


GongWilliam:
good job

PR #1167, opened 2021-06-21
fix: raft's vote don't increase the term

GongWilliam:
raft don't allow to increase the term number when it's in the candidate phase, it only can increase the term number when it's in the electionACK phase.

### DIFF
--- a/vendor/github.com/tiglabs/raft/raft_fsm.go
+++ b/vendor/github.com/tiglabs/raft/raft_fsm.go
@@ -20,9 +20,10 @@ import (
 	"math/rand"
 	"strings"
 
+	"time"
+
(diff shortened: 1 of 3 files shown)


shuoranliu (commented):
According to the Raft consensus algorithm, "terms" indentify obsolete information. But this commit seems to break the rule. Maybe we shall be more careful about such changes, and think twice.

GongWilliam:
> According to the Raft consensus algorithm, "terms" indentify obsolete information. But this commit seems to break the rule. Maybe we shall be more careful about such changes, and think twice.

this patch is a system optimization at the election stage, and it doesn't break the rule of raft's basic running.

PR #1236, opened 2021-10-09
[RFC] metanode: Load snapshot file with multiple goroutines

shyodx:
This patchset introduces a new format of snapshot files in metanode.
In order to distinguish between the new and old format, the new one
is called small file while the old one is called large file.

Small file is composed of fixed-size blocks, in which continous
entries are saved. If an entry is larger than a block, the entry is
saved to large file.

When metanode loads snapshot file during booting, multiple routines
load blocks of small file concurrency to reduce boot time.

cfs-cli is also adjusted to enable small file when creating a new
volume. Otherwise, small file is disabled by default.

The performance is tested with a large snapshot/inode file, which
has 133333338 Inode entries.

Without small file:

| snapshot size | read block size | load time |
| ------------- | --------------- | --------- |
| 12G           | 4M              | 1m35.238  |
| 12G           | 32M             | 1m35.470  |
| 12G           | 64M             | 1m32.916  |

With small file:

| snapshot size | read block size | routines | load time |
| --------------| --------------- | -------- | --------- |
| 12G           | 4M              | 128      | 49.574s   |
| 12G           | 4M              | 64       | 47.502s   |
| 12G           | 4M              | 16       | 48.353s   |
| 12G           | 4M              | 32       | 49.772s   |
| 12G           | 32M             | 32       | 49.252s   |
| 12G           | 64M             | 32       | 54.584s   |

### DIFF
--- a/cli/cmd/const.go
+++ b/cli/cmd/const.go
@@ -73,6 +73,7 @@ const (
 	CliFlagDelWorkerSleepMs   = "delete-worker-sleep-ms"
 	CliFlagMarkDelRate        = "mark-delete-rate"
 	CliFlagCrossZone          = "crossZone"
+	CliFlagMetaFileBlock      = "mf-blksize"
 
 	//CliFlagSetDataPartitionCount	= "count" use dp-count instead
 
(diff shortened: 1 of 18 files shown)


true1064:
The modification is very large
If there is a strong need for this optimization in the future, open it again

PR #1266, opened 2021-11-18
Fix meta resp and make fuse ReadDir compatible

shyodx:
1. metanode should give a response to client if any error occurs.
2. For old version metanode, ReadDirLimit may not implemented. ReadDir
   should have the capability to detect and rollback to ReadDirAll.

### DIFF
--- a/client/fs/dir.go
+++ b/client/fs/dir.go
@@ -297,9 +297,29 @@ func (d *Dir) ReadDir(ctx context.Context, req *fuse.ReadRequest, resp *fuse.Rea
 	start := time.Now()
 
 	dirCtx := d.dctx.GetCopy(req.Handle)
+	if d.super.noReaddirLimit {
+		if dirCtx.Name != "" {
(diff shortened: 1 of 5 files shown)


leonrayang:
Thanks @shyodx for these client/metanode robustness fixes!

We’re closing this PR because it has been open since 2021 and was never merged. On current master:

Unknown opcode handling — metanode still returns an error internally without sending a response packet, which can cause clients to wait until timeout (the issue your testcase demonstrated).
ReadDir compatibility — the fuse client always uses ReadDirLimit_ll and does not fall back to ReadDirAll when the metanode does not support limited readdir.
OpIntraGroupNetErr exists in proto but is not wired up for this case. If these issues still matter, please open a focused new PR (the unknown-opcode response fix is a small, high-value change). The ReadDir fallback may be lower priority given version skew today, but worth noting if mixed-version clusters are still supported.

Closing as stale. Would welcome a fresh PR for the opcode response fix!

PR #1320, opened 2022-01-14
Fuse clisent live upgrade

shyodx:
The patches allows to upgrade fuse client without umounting. The following
stes are executed to achieve that:
  1. stop reading more requests from /dev/fuse
  2. save context
     Context saves information of inodes/files, which are still in use (open
     or not evicted) between fuse client and fuse kernel module.
  3. save file descriptor of /dev/fuse
     Use Unix Domain Socket to transmit fd.
  4. old client exits and new client starts
  5. restore context
     New client tries to restore context instead of really mounting fuse.
  6. restore file descriptor of /dev/fuse
     New client keep reading pending requests.


shuoranliu (approved):

shuoranliu (approved):

Huweicai:
FYI
图片

PR #1379, opened 2022-03-18
CI: Fix large chunk of whitespace in ltp result

shyodx:
There are lots of large chunk of whitespace in ltp result. That
makes the result output diffault to read. The whitespace is caused
by truncating /tmp/ltp.log and redirecting it to a temporary file.

This patch simplies the execution of runltp and simplies the checking
of ltp log file.


shyodx:
Before:
before-1

After:
after

PR #1382, opened 2022-03-24
metanode: Adjust the usage of Btree.GetTree

shyodx:
Btree.GetTree() starts a COW clone for a tree. It would be better
not to clone a tree in the following scenarios:
  * when reporting the number of inodes or dentries, clone a tree
    does not help report a more precise value
  * if a tree is already cloned, do not clone it once more to get
    its length

This patch also removes duplicated definitions of getting inode
and dentry trees.

### DIFF
--- a/cli/cmd/compatibility.go
+++ b/cli/cmd/compatibility.go
@@ -102,7 +102,9 @@ func verifyDentry(client *api.MetaHttpClient, mp metanode.MetaPartition) (err er
 	if err != nil {
 		return
 	}
-	mp.CloneDentryTree().Ascend(func(d metanode.BtreeItem) bool {
+
+	cnt := 0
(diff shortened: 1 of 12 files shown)


leonrayang:
Thanks @shyodx for this btree usage cleanup!

We're closing this PR because it has been open since 2022 with unresolved conflicts and was never merged. On current master, BTree.GetTree() still performs a COW clone under that name, and the proposed CloneTree rename plus reduced cloning for count/reporting paths was not integrated.

Metanode storage has evolved considerably since then (RocksDB mode, MemSnapShot, etc.). If unnecessary btree cloning is still a performance concern, please open a fresh issue/PR against current master with benchmarks on the specific hot paths.

Closing as stale. Contributions welcome on a new PR!

PR #1385, opened 2022-03-25
cli: Fix failure of unmarshal Inode.Extents

shyodx:
When executing:
	cfs-cli compatibility meta ./ 192.168.0.22:17220 147

The following error returned:
```
[Meta partition is 147, verify result]
The number of dentry is 0, all dentry are consistent
Error: json: cannot unmarshal object into Go struct field Inode.Extents of type []proto.ExtentKey
```

The reason is that the definitions of `Extents' of `Inode' in cli and
metanode are different. The former is defined as a slice, while the
latter is defined as a non-exportable pointer. So ignore the key in
both marshal and unmarshal sides.

### DIFF
--- a/cli/api/metaapi.go
+++ b/cli/api/metaapi.go
@@ -25,10 +25,11 @@ import (
 	"time"
 
 	"bufio"
+	"io"
+
 	"github.com/cubefs/cubefs/metanode"
 	"github.com/cubefs/cubefs/proto"
 	"github.com/cubefs/cubefs/util/log"
-	"io"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


shyodx:
The verify logic of `verifyInode/verifyDentry' seems still not complete.
`inodeTree/dentryTree` is loaded from a directory which is specified by the parameter 'snpashotPath'. However, if the parameter is wrong, we can only get an empty `inodeTree/dentryTree`. Then, `verifyInode/verifyDentry` skips comparing the tree with data returned by `getAllInodes/getAllDentriy`.
Shall we fix this behavor? or it is designed that way?

PR #1401, opened 2022-04-08
 v3.0.0 support mulit layer data layer such as blobstore  (Erasure-Code) and other features

leonrayang:
This is a very important pr with a lot of code to support mulit layer data layer such as blobstore (Erasure-Code)
Before this we released v3.0.0-beta(https://github.com/cubeFS/cubefs/releases/tag/v3.0.0-beta) which make volume can be create as standard(3 replica) or low frequency access type(Erasure-Code) but not merged into master, and we introduce a new module BlobStore subsystem which supports erasure coding storage method(https://github.com/cubeFS/cubefs-blobstore) to support low frequency volume and reduce cost.

this pr also include main features as 
1) BlobStore subsystem which supports erasure coding storage.
2) multicache layer, 1st level:client local read cache; 2nd level:hot pool replica; 3nd level:code pool based on erasure code.
3) single data replica supported.

### DIFF
--- a/client/fuse.go
+++ b/client/fuse.go
@@ -559,6 +559,7 @@ func mount(opt *proto.MountOptions) (fsConn *fuse.Conn, super *cfs.Super, err er
 
 	if opt.EnablePosixACL {
 		options = append(options, fuse.PosixACL())
+		options = append(options, fuse.DefaultPermissions())
 	}
 
 	if opt.EnableUnixPermission {


hooklee2000:
ci-tests failed for missing cmake when build snappy. Maybe need to  apt install cmake in docker/[Dockerfile](https://github.com/cubeFS/cubefs/blob/master/docker/Dockerfile)

leonrayang:
> 

we did some modiftion on third part libraray in the past which need move out from vendor otherwise will be coverd by go mod and load from github. The unit test and shell script need some modifition.

PR #1587, opened 2022-10-31
FIX: fix the metrics of volume space

liubingxing:
Signed-off-by: liubingxing <liubbingxing@gmail.com>

we create a volume with 10000GB, but the metrics show the wrong result.
image

image

The volume metrics: **cfs_master_vol_total_GB** and **cfs_master_vol_used_GB** use **volStatInfo.TotalSize** which count by **bytes**.
We should fix the metrics by **volStatInfo.TotalSize/util.GB**


leonrayang:
cli volume info use master's vol.Capacity directly and should not need to divide by GB

liubingxing:
@leonrayang The cli volume info is right, but the volume info from http metrics need to fix.

Victor1319 (dismissed):

Victor1319:
merge  after grafana dashboard template json file also be updated, to avoid users confused when using metrics monitor

leonrayang:
> merge after grafana dashboard template json file also be updated, to avoid users confused when using metrics monitor

ok

true1064:
@liubingxing please resolve the confilicts

sejust (approved):

sejust:
move to #2800

PR #1598, opened 2022-11-04
Enhancement: change the DentryCache sync.Mutex to sync.RWMutex

liubingxing:
Signed-off-by: liubingxing <liubbingxing@gmail.com>

Change the DentryCache sync.Mutex to sync.RWMutex to enhance the Concurrent call of DentryCache.Get()


liubingxing:
since the DentryCache.Get() will change the cache item, so it must use sync.Mutex
image

PR #1640, opened 2022-12-28
[Enhancemant] Demand write nodes should be calculated based on zone number and replica number

tomscut:

Demand write nodes should be calculated based on zone number and replica number.  For example, if the number of zones is greater than the number of replicas, the number of demand write nodes is 1 will be ok.  Currently, a volume(`crosszone=true`) cannot be created when there is only one metanode in each zone.

### DIFF
--- a/master/topology.go
+++ b/master/topology.go
@@ -1071,12 +1071,10 @@ func (t *topology) getZoneByIndex(index int) (zone *Zone) {
 func calculateDemandWriteNodes(zoneNum int, replicaNum int) (demandWriteNodes int) {
 	if zoneNum == 1 {
 		demandWriteNodes = replicaNum
+	} else if zoneNum >= replicaNum {
+		demandWriteNodes = 1
 	} else {
-		if replicaNum == 1 {
-			demandWriteNodes = 1
-		} else {
-			demandWriteNodes = 2
-		}
+		demandWriteNodes = 2
 	}
 	return
 }


leonrayang:
> Signed-off-by: Tao Li [tomscut@apache.org](mailto:tomscut@apache.org)
> 
> **What this PR does / why we need it**: Demand write nodes should be calculated based on zone number and replica number. For example, if the number of zones is greater than the number of replicas, the number of demand write nodes is 1 will be ok. Currently, a volume(`crosszone=true`) cannot be created when there is only one metanode in each zone.

The allocation policy considered the client cache , which include the data partitions info, which be updated peridically 1 minites,  there will no usable data partition if the zone cann't be usable，because every partition be affected. But when the partitons allocated as  2+1 while there's 3 zones, that means there's still usable dataparititons in client's cache if one zone is unavialbe.

Cloudstriff:
您好，您的邮件已收到。

leonrayang:
Thanks @tomscut for this enhancement proposal!

We're closing this PR because it has been open since late 2022, was labeled INVALID, and was never merged. The proposed change — relaxing per-zone writable node requirements when the number of zones exceeds the replica count — was discussed but not accepted. The concern raised during review is that client data-partition caches depend on having sufficient writable nodes per zone to maintain usable partitions during zone failures (e.g. 2+1 replication across 3 zones).

On current master, calculateDemandWriteNodes() still uses the original conservative logic. If cross-zone volume creation with minimal per-zone metanode counts remains a requirement, please open a new issue describing the deployment scenario so we can evaluate alternative approaches (e.g. partition allocation policy changes) rather than only relaxing the writable-node threshold.

Closing as stale. Thank you for the contribution!

PR #1644, opened 2023-01-05
[Enhancement] Update the NodeView for displaying status

tomscut:

Update the `NodeView` for displaying status. `Status:true` -> `Active:true`.
image


leonrayang:
Thanks @tomscut for this NodeView clarity improvement!

We're closing this PR because it has been open since 2023 with a DO NOT MERGE label and was never integrated. On current master, proto.NodeView still uses the Status field (populated from isActive), not Active.

While renaming Status to Active would improve readability, it is a breaking API change for cfs-cli, monitoring dashboards, and other consumers. If this is still desired, please open a new issue/PR that includes a migration plan (e.g. deprecate Status while introducing Active, or document the breaking change in release notes).

Closing as stale. Thank you!

PR #1674, opened 2023-02-17
[Enhancemant] Support fuse request timeout, add client metrics

tomscut:

A fuse client can hang due to a server exception or hardware failure. We have implemented a timeout mechanism for fuse clients, which is not enabled by default but is enabled by configuring `requestTimeout`. Currently, this feature works well in our cluster.

### DIFF
--- a/client/fuse.go
+++ b/client/fuse.go
@@ -447,7 +447,7 @@ func main() {
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


leonrayang (dismissed):
need resolve confilicts

true1064 (dismissed):

leonrayang (approved):

PR #1697, opened 2023-02-28
Fix copyobject test

tangdeyi:
[CI] fix copyobject testcase


sejust:
@tangdeyi  pls pass the DCO checking

PR #1739, opened 2023-03-05
[Feature] add file statistics function

liubingxing:
Adding the file stats function to analyse the files for each volume. 
It will read the inode file from snapshot dir and count the files for several size range.
We can view the file distribution for each volume on grafana.
image

We also add a cmd to control the file stats function enable, the default value is false.

### DIFF
--- a/metanode/partition_file_stats.go
+++ b/metanode/partition_file_stats.go
@@ -1,0 +1,136 @@
+package metanode
+
+import (
+	"fmt"
+	"github.com/cubefs/cubefs/proto"
+	"github.com/cubefs/cubefs/util/exporter"
+	"time"
+)
+
+type FileSizeRange uint32
+
+const (
+	Size1K   uint64 = 2 << 10
+	Size1M   uint64 = 2 << 20
+	Size16M         = 16 * Size1M
+	Size32M         = 32 * Size1M
+	Size64M         = 64 * Size1M
+	Size128M        = 128 * Size1M
+	Size256M        = 256 * Size1M
+)
+
+const (
+	LessThan1K FileSizeRange = iota
+	LessThan1M
+	LessThan16M
+	LessThan32M
+	LessThan64M
+	LessThan128M
+	LessThan256M
+	BiggerThan256M
+	MaxRangeType
+)
+
+const (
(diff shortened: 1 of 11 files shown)


metanode/partition_file_stats.go
```diff
@@ -0,0 +1,215 @@
...
+		// sleep a while if iter StatsInterval count
+		if count > StatsInterval {
+			count = 0
+			time.Sleep(StatsSleep)
+		}
+	}
+}
+
+func (mp *metaPartition) startFileStats() {
+	checkTicker := time.NewTicker(fileStatsCheckPeriod)
```
leonrayang, line 172: 10s interval may cause cpu cost too much

liubingxing: The **fileStatsCheckPeriod** here is mainly used to detect the status change of mp (such as leader to follower). The period for **fileStats** is **fileStatsHandlePeriod** (default value is 15min).

metanode/partition_file_stats.go
```diff
@@ -0,0 +1,215 @@
...
+		labels := map[string]string{
+			"partid":    fmt.Sprintf("%d", mp.config.PartitionId),
+			"volName":   mp.config.VolName,
+			"sizeRange": toString(FileSizeRange(i)),
+		}
+		exporter.NewGauge("fileStats").SetWithLabels(float64(val), labels)
+	}
+}
+
+func (mp *metaPartition) fileStats(rootDir string) {
```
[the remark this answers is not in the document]
liubingxing: > recommend use mp snapshot store process to get all inodes info, the io cost seems not nessesary

That's good advice, and I'll optimize the code and test it again later.

liubingxing:
@leonrayang This pr has been updated with the following changes. Please review again, thank you very much.
1. Collect inodes information during the processing of storing mp snapshot;
2. Adjust the check interval from 10s to 30s;

leonrayang:
@M1eyu2018  gofmt not pass

liubingxing:
@leonrayang The ci-tests process is a bit abnormal, and the following files are not subject to this pr changes.
image

leonrayang (approved):

PR #1800, opened 2023-03-15
docs(docs-zh): add document capacity.md in dir maintenance

true1064:
docs(docs-zh): add document capacity.md in dir maintenance


docs-zh/source/maintenance/capacity.md
```diff
@@ -0,0 +1,26 @@
+容量管理
```
pengtianyue025, line 1: 这个格式好像不是md格式的

true1064: the title is fixed

PR #1862, opened 2023-03-27
 Feature: add ClusterUUID for mn/dn to identify a cluster upon restart

liubingxing:
Currently, metanode/datanode may be added to other clusters due to the incorrect configuration during mn/dn restart.
If more than two metanode are mistakenly added to other clusters, some metapartition may be missing two replicas, resulting in client access errors.

We add the cluster identification function, which generates the unique cluster ID through commands and persists it in the master rocksdb. When metanode and datanode start for the first time, the ID will be obtained from the master and persists to the data directory, and will be checked when the metanode and datanode restart. 

We add a few new commands in this pr:
1. Command to generate uuid;
2. Command to query clusterUUID;
3. Command to enable metanode/datanode to check clusterUUID during startup, the default value is false;
4. Command to generate CLUSTER-VERSION file for metanode (used in the old cluster）
5. Command to generate CLUSTER-VERSION file for datanode (used in the old cluster）

For the newly cluster, we can use the following steps to enable clusterUUID function:
1. After all master startup, using the command to generate clusterUUID;
2. Enable the clusterUUID check function;
3. Start metanode and datanode;

If you want to upgrade this function in an old cluster
1. Generate clusterUUID;
2. Restart metanode with the new code;
3. Use the command to generate CLUSTER-VERSION for metanode;
6. Restart datanode with the new code;
7. Use the command to generate CLUSTER-VERSION for datanode;
8. Enable the clusterUUID check function;

### DIFF
--- a/datanode/partition_raft.go
+++ b/datanode/partition_raft.go
@@ -515,6 +515,13 @@ func (s *DataNode) startRaftServer(cfg *config.Config) (err error) {
 
 	s.parseRaftConfig(cfg)
 
+	if s.clusterUuidEnable {
+		if err = config.CheckOrStoreClusterUuid(s.raftDir, s.clusterUuid, false); err != nil {
+			log.LogErrorf("CheckOrStoreClusterUuid failed: %v", err)
+			return fmt.Errorf("CheckOrStoreClusterUuid failed: %v", err)
+		}
+	}
+
 	constCfg := config.ConstConfig{
 		Listen:           s.port,
 		RaftHeartbetPort: s.raftHeartbeat,
--- a/datanode/server.go
+++ b/datanode/server.go
@@ -148,6 +148,8 @@ type DataNode struct {
 	diskIopsWriteLimit      uint64
 	diskFlowReadLimit       uint64
 	diskFlowWriteLimit      uint64
+	clusterUuid             string
+	clusterUuidEnable       bool
 }
 
 func NewServer() *DataNode {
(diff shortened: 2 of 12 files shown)


leonrayang:
@liubingxing，thank you for you pull request. I din't under the the third part of your comment. Why the old cluster need create the CLUSTER-VERSION by hand? When update the new release version, first upate master and create uuid, then the metaNode and dataNode updated and will load from the master with the cluster id, and will create the CLUSTER-VERSION. No matter what, the CLUSTER-VERSION be create by hand is not accepted.

liubingxing:
> @liubingxing，thank you for you pull request. I din't under the the third part of your comment. Why the old cluster need create the CLUSTER-VERSION by hand? When update the new release version, first upate master and create uuid, then the metaNode and dataNode updated and will load from the master with the cluster id, and will create the CLUSTER-VERSION. No matter what, the CLUSTER-VERSION be create by hand is not accepted.

@leonrayang Sorry for taking so long to reply.
The reason why mn/dn in the old cluster do not automatically generate CLUSTER-VERSION upon restart is that considering the following scenarios.
If the UUID function is enabled in cluster A but not in cluster B, and the mn/dn in cluster B uses the wrong configuration of cluster A during the upgrade. The CLUSTER-VERSION will be generated and the node will be added to cluster A.
To avoid this error, we believe that the nodes in the old cluster should be identified and generated CLUSTER-VERSION by the administrator.
Choosing to copy the configuration file by hand does cause some inconvenience, we will modify it as soon as possible.

liubingxing:
@leonrayang 
I have updated the code and comment, please review again, thanks a lot.

PR #1950, opened 2023-05-06
[Enhancement]Add faq & customize CXXFLAGS support on building `rocksdb`

cloorc:

### To make CXXFLAGS customizable to prevent failure from `cc1plus: all warnings are treated as errors` on building `rocksdb`

To make it effective, just add `export CXXFLAGS=-Wno-error` at the end of `env.sh` before executing `source env.sh`.

**Release note**:

```release-note
```

### DIFF
--- a/docs-zh/source/faq/build.md
+++ b/docs-zh/source/faq/build.md
@@ -49,6 +49,8 @@ export DISABLE_WARNING_AS_ERROR=true
 
 - 对于`cubefs`根目录而言，请在根目录下的`env.sh`中添加导出环境变量：
 
+> **注意**： 选项开关可能因gcc版本不同存在差异，以下选项在`gcc (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0`版本上测试通过
+
 ```bash
 export CXXFLAGS=-Wno-error
 ```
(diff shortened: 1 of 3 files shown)


docs-zh/source/faq/build.md
```diff
@@ -23,18 +23,32 @@
...
+- 对于`blobstore`模块来说，进入目录`blobstore`后，在`blobstore/env.sh`文件中添加以下新的环境变量后执行`source env.sh`后继续编译:
+
+```bash
+export DISABLE_WARNING_AS_ERROR=true
+```
+
+- 对于`cubefs`根目录而言，请在根目录下的`env.sh`中添加导出环境变量：
+
+```bash
+export CXXFLAGS=-Wno-error
```
leonrayang, line 53: hi,cloorc, I am not sure the flag wno-error could be supported in different gcc version, as https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65403

cloorc: Hmm, it's might be version dependent. What if adding a specific gcc version as extra condition? Of course, flags here are optional. Develoeprs might add whatever options accordingly. And it works in my circumstance with specific gcc version.

leonrayang: > Hmm, it's might be version dependent. What if adding a specific gcc version as extra condition? Of course, flags here are optional. Develoeprs might add whatever options accordingly. And it works in my circumstance with specific gcc version.

I consider "flags here are optional" should make a declaration. or else users may make a mistake

cloorc: Ok, I've updated related description, please have a look again. Thanks!

PR #1952, opened 2023-05-06
[Enhancement] metanode: GetCurrentTimeUnix use atomic

sjp00556:
GetCurrentTimeUnix use atomic is much quicker

**benchmark**
```
$ go test -bench="^BenchmarkGet" -run=none
goos: linux
goarch: amd64
pkg: github.com/cubefs/cubefs/metanode
cpu: Intel(R) Core(TM) i7-10700 CPU @ 2.90GHz
BenchmarkGetCurrentTimeUnix-16    	1000000000	         0.3317 ns/op
BenchmarkGetCurrentTime-16        	100000000	        10.98 ns/op
BenchmarkGetNowTime-16            	43397809	        27.76 ns/op
PASS
ok  	github.com/cubefs/cubefs/metanode	2.796s
```

**Release note**:

```release-note
```

### DIFF
--- a/metanode/time_test.go
+++ b/metanode/time_test.go
@@ -1,0 +1,41 @@
+package metanode
+
+import (
+	"testing"
+	"time"
+)
+
+func TestGetCurrentTimeUnix(t *testing.T) {
+	t1 := Now.GetCurrentTimeUnix()
+	t2 := Now.GetCurrentTime().Unix()
+	n := time.Now().Unix()
+	if n-t1 > 1 || n-t2 > 1 {
+		t.Errorf("wrong time: %d %d %d", n, t1, t2)
+	}
+
+	time.Sleep(time.Second * 2)
+	t1 = Now.GetCurrentTimeUnix()
+	t2 = Now.GetCurrentTime().Unix()
+	n = time.Now().Unix()
+	if n-t1 > 1 || n-t2 > 1 {
+		t.Errorf("wrong time: %d %d %d", n, t1, t2)
+	}
+}
+
+func BenchmarkGetCurrentTimeUnix(b *testing.B) {
+	for i := 0; i < b.N; i++ {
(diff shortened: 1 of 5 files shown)


metanode/time_test.go
```diff
@@ -0,0 +1,41 @@
+package metanode
```
sejust, line 1: Add the "License" into head of the file.

sejust (changes requested):

sjp00556:
> @sjp00556 Hello, you improve benchmark by adding new a function. It's great, but not the better way in this case.
> 
> 1. without changing the effect of the original function，change `sync.RWMutex` to `atomic.Value`.
> 2. changing return type of function `GetCurrentTime`, just store unix timestamp into `NowTime`.

@sejust maybe some situations need time.Time usage like this: Now.GetCurrentTime().UnixMilli() in future, so i kept GetCurrentTime func and add GetCurrentTimeUnix func which is only to get time unix.

sejust:
> 1. atomic.Value



> 1. without changing the effect of the original function，change `sync.RWMutex` to `atomic.Value`



> > @sjp00556 Hello, you improve benchmark by adding new a function. It's great, but not the better way in this case.
> > 
> > 1. without changing the effect of the original function，change `sync.RWMutex` to `atomic.Value`.
> > 2. changing return type of function `GetCurrentTime`, just store unix timestamp into `NowTime`.
> 
> @sejust maybe some situations need time.Time usage like this: Now.GetCurrentTime().UnixMilli() in future, so i kept GetCurrentTime func and add GetCurrentTimeUnix func which is only to get time unix and for benchmark purpose

ok，store `now(time.Time)` into atomic.Value without adding a new function.

leonrayang (approved):

leonrayang (approved):

PR #1956, opened 2023-05-09
[Fix] client selects local dp

M1eyu2018:
[Fix] client selects local dp


M1eyu2018:
https://github.com/cubefs/cubefs/pull/1958

Issue #1904, opened 2023-04-17
[Bug]: Client mount point supports non-repeatable mount


pengtianyue025:
**Describe the bug**
If the client starts multiple times, the mount point will be mounted repeatedly.

**Expected behavior**
Client mount point supports non-repeatable mount.

PR #1961, opened 2023-05-10
mount point supports non-repeatable mount

mfordjody:
Signed-off-by: mfordjody [mfordjody@gmail.com](mailto:mfordjody@gmail.com)

Client mount point supports non-repeatable mount.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes https://github.com/cubefs/cubefs/issues/1904

None

**Release note**:

```release-note
None
```

### DIFF
--- a/depends/jacobsa/fuse/mount.go
+++ b/depends/jacobsa/fuse/mount.go
@@ -23,7 +23,7 @@ import (
 // Server is an interface for any type that knows how to serve ops read from a
 // connection.
 type Server interface {
-	// Read and serve ops from the supplied connection until EOF. Do not return
+	// ServeOps Read and serve ops from the supplied connection until EOF. Do not return
 	// until all operations have been responded to. Must not be called more than
 	// once.
 	ServeOps(*Connection)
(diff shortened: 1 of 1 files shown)


leonrayang:
@mfordjody Can you explain in detail the principle of your implementation?

mfordjody:
> @mfordjody Can you explain in detail the principle of your implementation?

If the length of the `dir` string is 1, then it is returned immediately. It is also returned directly if the `OpContext` field of the `config` structure is not `nil`. Send a `nil` value to `ready` to notify that the mount operation is ready. Open the `/dev/' directory with `os.Open` and return it directly if an error occurs during the open operation, or if the error indicates that the file or directory already exists. Finally, at the end of the function execution, the file descriptor `dev` is closed and the corresponding resources are freed.

mfordjody:
> @mfordjody能详细说说你的实现原理吗？

If the length of the dir string is 1, then it will be returned immediately. It is also returned directly if the OpContext field of the config structure is not nil. Send a nil value to ready to notify that the mount operation is ready. Open the /dev/ directory with os.Open and return directly if an error occurs during the open operation, or if the error indicates that the file or directory already exists. Finally, at the end of the function execution, the file descriptor dev is closed and the corresponding resources are freed.

leonrayang:
Not suitable for merging

PR #2027, opened 2023-05-29
build: put rocksdb libs into the docker image

cessory:
put rocksdb's libs into the docker image, avoid download file error


pengtianyue025 (approved):

pengtianyue025:
nice

PR #2031, opened 2023-05-29
enhance(client): improve the client restore process

liubingxing:
The following figure shows the process of upgrading cfs-client. This is a good feature and we want to use it in production environment. 
We tested it for a while and found some problems, especially when the old client processing quit but the new client failed to start up, will cause the test failed.
image

So we want to improve the process that the old client waits until the new client restored data.
image

The main changes are as follow:
1. Restoring the nodecache correctly, including the name attribute
2. The old client continues to suspend after its trySuspend function is completed. After the new client restored data, it sends the shutdown command to make the old client quit.

### DIFF
--- a/client/fs/dir.go
+++ b/client/fs/dir.go
@@ -140,6 +140,7 @@ func (d *Dir) Attr(ctx context.Context, a *fuse.Attr) error {
 		return ParseError(err)
 	}
 	fillAttr(info, a)
+	a.ParentIno = d.parentIno
 	log.LogDebugf("TRACE Attr: inode(%v)", info)
 	return nil
 }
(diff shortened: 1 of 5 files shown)


liubingxing:
@leonrayang @Victor1319 Please take a look at this PR. Thanks a lot.

bboyCH4 (dismissed):

liubingxing:
@sejust Code conflict resolved, please review again. Thanks a lot.

leonrayang:
Thanks @liubingxing for improving the cfs-client hot-upgrade restore flow!

We're closing this PR because it has been open since 2023 and was never merged, despite earlier review interest. Basic FUSE suspend/restore (-r option) exists on current master, but the proposed improvements — coordinated old/new client handoff, nodecache name restoration, and parentIno in dir.Attr — were not integrated.

If production hot-upgrade reliability is still a concern, please open a fresh PR against current master with a test plan covering the failure scenario (old client exits but new client fails to start).

Closing as stale. Would welcome an updated PR!

PR #2047, opened 2023-05-31
chore(README): adjust chapter order 

xiaochunhe:

Adjust the position of the 'contributors' section to avoid interrupting the coherence of the document content


pengtianyue025:
The DCO check is failed, maybe you can fix it.

PR #2060, opened 2023-06-02
Blobstore v1.2.2

pengtianyue025:

- support consume group
- support 3-az module register to consul
- update initial volume allocation logic


sejust (approved):
Good!

Issue #1919, opened 2023-04-17
[Feature]: try to reduce load of datanode and metanode


pengtianyue025:
**Describe the solution you'd like**
too many partitions may cost a lot cpu, io resource, that cause node's load high, and become slow, try to reduce load.

**Additional context**

setcy:
Hello @Victor1319, I encountered some questions while processing the issue.

Does the master need to notify DataNodes and MetaNodes to stop activities on their partitions? If the answer is yes, it will lead to two ways of stopping: synchronous and asynchronous. Below, I will explain my understanding of both methods.

### Synchronous stopping

Synchronous stopping is difficult to handle in cases of partial errors. For example, if the master notifies the DataNode to stop successfully, but an error occurs when notifying the MetaNode, how should it be handled? If a rollback is chosen, what happens if an error occurs during the rollback process?

### Asynchronous stopping

Asynchronous stopping is more complex. Three additional status, "markSleep", "sleep", and "markNormal", need to be added to the volume. A scheduled task is required to handle volumes marked as "markSleep" and "markNormal", and once all tasks are completed, they are re-marked as sleep and normal. I prefer this approach, but is there a better solution?

Victor1319:
Hello @setcy, after careful consideration about this issue, I feel that if you are not very familiar with the cubefs system, it is too difficult to realize the ability to hibernate. 
You need to consider the state of the volume, the state of the partition, the interaction with each node, and the migration process.
The **main problem** of this issue is that when there are too many partition, the node load is high, I suggest that you can consider how to reduce the load.
For example, you can increase the interval of some scheduled tasks (configurable), and the start time of scheduled tasks is more dispersed, tasks eg: `dp size comute`,  `mp snapshot`, `dp crc compute`, and so on.
If you want to implement it in this way, please tell me, and I would adjust the title of the issue for you

setcy:
> For example, you can increase the interval of some scheduled tasks (configurable), and the start time of scheduled tasks is more dispersed, tasks eg: `dp size comute`, `mp snapshot`, `dp crc compute`, and so on.
OK, I will try to implement it in this way. Please adjust the title for me, thank you.

setcy:
[...]

PR #2068, opened 2023-06-03
feat: implement configurable interval and dispersed start time for mpPersistDataInternal

setcy:

When there are too many inodes and dentries in meta partition, this schuedule will cause high cpu costs.
This issue is going to implement an ability to change the mpPersistDataInternal, to lessen cpu costs.

- implemented the feature to introduce a randomized delay of 0 to 30 seconds for the scheduled task
- can configure the partition scheduled task time when metanode starts by editing the configuration file
- When creating a partition, the first execution time will take a random value between [0, persistence time] to ensure that execution does not start simultaneously

We now can use `cfs-cli metanode info [ADDRESS]` to see the current persistDataInternal

### before the implementation
image

### after the implementation, and use the default value 300 (5min)
image

### after the implementation, and set the value to 1200 (20min)
image

Which issue this PR fixes
fixes https://github.com/cubefs/cubefs/issues/1919

### DIFF
--- a/metanode/partition_store_ticket.go
+++ b/metanode/partition_store_ticket.go
@@ -172,8 +172,9 @@ func resetTimer(timer *time.Timer, ts int64) {
 	timer.Reset(time.Duration(ts+r.Int63n(RandomIntervalS)) * time.Second)
 }
 
-// Avoid devices starting scheduled tasks simultaneously
+// resetTimerWithRandomTime resets the timer with a random duration in max*[1, 2)
+// to avoid devices starting scheduled tasks simultaneously.
 func resetTimerWithRandomTime(timer *time.Timer, maxts int64) {
 	r := rand.New(rand.NewSource(time.Now().UnixNano()))
-	timer.Reset(time.Duration(float32(maxts)*r.Float32()) * time.Second)
+	timer.Reset(time.Duration(maxts+int64(float32(maxts)*r.Float32())) * time.Second)
 }
--- a/metanode/partition_store_ticket.go
+++ b/metanode/partition_store_ticket.go
@@ -16,6 +16,7 @@ package metanode
 
 import (
 	"encoding/binary"
+	"math/rand"
 	"sync/atomic"
 	"time"
 
@@ -24,6 +25,8 @@ import (
 	"github.com/cubefs/cubefs/util/exporter"
 	"github.com/cubefs/cubefs/util/log"
 )
+
+const RandomIntervalS = 30
 
 type storeMsg struct {
 	command       uint32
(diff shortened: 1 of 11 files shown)


Victor1319:
hello @setcy , there are some places that need to be adjusted, the job can be more simple.
1. Periodic persistence time can be made configurable. There is no need to set one for each volume. If there are many volumes, it will be more troublesome to set up. It can be configured in the metanode configuration file Enough
2. Regarding the randomization of periodic persistence, you can consider any value between sleep between [0, persistence time] when go func(stopC chan bool) {} run in startSchedule starts, so that it will be more random,
3. You can post the cpu graph before and after the implementation to see the effect

setcy:
> hello @setcy , there are some places that need to be adjusted, the job can be more simple.
> 
> 1. Periodic persistence time can be made configurable. There is no need to set one for each volume. If there are many volumes, it will be more troublesome to set up. It can be configured in the metanode configuration file Enough
> 2. Regarding the randomization of periodic persistence, you can consider any value between sleep between [0, persistence time] when go func(stopC chan bool) {} run in startSchedule starts, so that it will be more random,
> 3. You can post the cpu graph before and after the implementation to see the effect

Hello @Victor1319, I have implemented all of them, please review again, thank you!

metanode/partition_store_ticket.go
```diff
@@ -162,3 +166,14 @@ func (mp *metaPartition) stop() {
...
+
+func resetTimer(timer *time.Timer, ts int64) {
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
+	timer.Reset(time.Duration(ts+r.Int63n(RandomIntervalS)) * time.Second)
+}
+
+// Avoid devices starting scheduled tasks simultaneously
+func resetTimerWithRandomTime(timer *time.Timer, maxts int64) {
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
+	timer.Reset(time.Duration(float32(maxts)*r.Float32()) * time.Second)
```
Victor1319, line 178: here is max*[0,1), it's better to update as max*[1, 2), to avoid snapshot too frequently, you can you max+r.rand(max)

Victor1319: and please to upate your graph after change.

setcy: Fixed. And the graph will coming soon.

setcy:
## before the implementation
{B8B03230-C5E0-4961-BFAD-15B276EA955C}
## after the implementation, and use the default value 300 (5min)
{56CB446B-514A-456f-8C70-9EED18B04BCA}

PR #2083, opened 2023-06-07
feat(master): the cmd 'cfs-cli datapartition check' not involve markD…

true1064:
feat(master): the cmd 'cfs-cli datapartition check' not involve markDelete volumes. 
if the cmd shows dp of markDelete vlumes, it may interferes with operation and maintenance personnel.


bboyCH4 (approved):

wuchunhuan (approved):
LGTM

PR #2099, opened 2023-06-08
enhance(cli,client):1. add check inode max mp check 2.client return err

bboyCH4:
enhance(cli,client):1. add check inode max mp check 2.client return err


leonrayang (approved):
LGTM

true1064 (approved):

Issue #2107, opened 2023-06-08
[Bug]: Quota dual directory configuration optimization problem


baijiaruo:
**Describe the bug**
1. Set up dual directories. If the second one is just ',', it will be used as the root directory, and there will be one more row when querying, but it is unrealistic. paht only has pid and riniode data, which needs to be optimized.
---This path is not very uniform, dir6 can be recognized without /. If it is just a comma, it will be recognized as the root directory. It is better to start from the root directory according to the unified
2. This prompt can be optimized. Add a better directory name. Which directory is repeated is more intuitive. This requires an additional query.This is not necessarily easy to do, the master side does not know the relationship between the directory and the inode, it is purely comparing the inodeId.The path where your master can find this piece of data is the directory. Why do you say you can’t find it?
3. When setting multiple directories, three or seven can be used. It is recommended to increase the upper limit to prevent misuse
4. 4 used files are displayed after the quota synchronization of 7 directories is created
5. Let’s see if this list can be sorted. Every search is in random order.

xiangcai1215:
which version？

PR #2108, opened 2023-06-08
fix(master,cli): Multi-directory quota problem optimization

baijiaruo:
close: #2107

    1.create quota dir without / without error.
    2.Create quota with the same directory error message optimization
    3.A single quota increases the directory limit to 5
    4.Creating a quota, the scene of multiple directories shows that the number of used files is incorrect
    5.list quota Sort by quotaId from small to large

### DIFF
--- a/cli/cmd/quota.go
+++ b/cli/cmd/quota.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"math"
 	"path/filepath"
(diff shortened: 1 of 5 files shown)


Victor1319 (approved):

leonrayang (approved):
LGTM

PR #2112, opened 2023-06-09
[Feature]: Master qosClientId allocation optimization

### LINKED ISSUE #1918: [Feature]: Master qosClientId allocation optimization
- It can be optimized to allocate a batch of id numbers at a time, such as 100, so that multiple requests can be persisted at one time, reducing raft and disk pressure.

**Additional context**
At present, when each client starts, it will apply for a clientId (self-increasing unique id) from the master for qos flow control, and the id will be persisted to rocksdb through raft every time it is applied. If there are multiple new client mount requests frequently, it will cause the raft log to grow too fast, and because it needs to be applied to the disk, there is a requirement for disk performance.

NaturalSelect:

* This pull request optimize the allocation of `qosClientId`.
* ~~The allocation of `qosClientId` doesn't require raft to persist data.~~
* ~~Compatibility with older versions of code.~~
* ~~Conflict check for low bits overflow is supported.~~

fixes #1918 

* ~~This pull request assuming it's impossible for `qosClientId` from `0` to grow to max value of `uint32` in one term.~~
* ~~This pull request assuming it's impossible for raft term grow to max value of `uint32`.~~
* Please contract me if this pull request have passed your review, then I will rebase and rearrange commit points.

**Release note**:

```release-note
```

### DIFF
--- a/master/id_allocator.go
+++ b/master/id_allocator.go
@@ -30,6 +30,7 @@ type IDAllocator struct {
 	metaPartitionID uint64
 	commonID        uint64
 	clientID        uint64
+	clientIDLimit   uint64
 	quotaID         uint32
 	store           *raftstore.RocksDBStore
 	partition       raftstore.Partition
@@ -38,6 +39,8 @@ type IDAllocator struct {
 	mnIDLock        sync.RWMutex
 	qaIDLock        sync.RWMutex
 }
+
+const clientIDBatchCount = 1000
 
 func newIDAllocator(store *raftstore.RocksDBStore, partition raftstore.Partition) (alloc *IDAllocator) {
 	alloc = new(IDAllocator)
@@ -50,26 +53,8 @@ func (alloc *IDAllocator) restore() {
 	alloc.restoreMaxDataPartitionID()
 	alloc.restoreMaxMetaPartitionID()
 	alloc.restoreMaxCommonID()
-	alloc.restoreClientID()
 	alloc.restoreMaxQuotaID()
-}
-
-func (alloc *IDAllocator) restoreClientID() {
-	value, err := alloc.store.Get(maxClientIDKey)
-	if err != nil {
-		panic(fmt.Sprintf("Failed to restore maxClientID,err:%v ", err.Error()))
-	}
-	bytes := value.([]byte)
-	if len(bytes) == 0 {
-		alloc.clientID = 0
-		return
-	}
-	clientID, err := strconv.ParseUint(string(bytes), 10, 64)
-	if err != nil {
-		panic(fmt.Sprintf("Failed to restore maxClientID,err:%v ", err.Error()))
-	}
-	alloc.clientID = clientID
-	log.LogInfof("action[restoreClientID] maxClientID[%v]", alloc.clientID)
+	alloc.restoreClientID()
 }
 
 func (alloc *IDAllocator) restoreMaxDataPartitionID() {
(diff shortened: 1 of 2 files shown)


master/id_allocator.go
```diff
@@ -216,21 +203,23 @@ errHandler:
...
+		// FIXME: give up current term
+		err = ErrClientIdRunOut
 		goto errHandler
 	}
-	alloc.setClientID(clientID)
+	counterPart += 1
+	clientID = term << 32
+	clientID |= uint64(counterPart)
+	alloc.clientCounter = counterPart
 	return
```
Victor1319, line 223: here, you can use `alloc.clientCounter ++`, no need to use a tmp variable counterPart

NaturalSelect: Yes, I will fix it

Victor1319 (dismissed):

Victor1319 (dismissed):

Victor1319:
LGTM

master/id_allocator.go
```diff
@@ -216,21 +203,20 @@ errHandler:
...
-	if err != nil {
+	var leader uint64
+	var term uint64
+	leader, term = alloc.partition.LeaderTerm()
+	if leader != alloc.partition.Status().NodeID {
 		goto errHandler
 	}
-	if _, err = alloc.partition.Submit(cmd); err != nil {
+	if alloc.clientCounter == math.MaxUint32 {
+		// FIXME: give up current term
```
leonrayang, line 213: Master-slave switching occurs during network jitter. In ideal network conditions, master-slave switching occurs rarely. However, if there are 500 ID requests per second for several months, it will overflow after more than three months of continuous operation, putting the risk of client ID conflicts.

NaturalSelect: If there is a possibility of conflicting ids(the allocated id reached the max value of `uint32` in a term), the master will return an error.

This is a temporary scheme.

If we can give up current term and switch to next term, this problem could be solved and the FIXME should be removed.

NaturalSelect: Also see special note for reviewers.

leonrayang: > Also see special note for reviewers.

We can add a check process after get the id and retry if exist,the confict is a rare accident

NaturalSelect: Good idea!

I will update this pull request tonight.

NaturalSelect: OK, this feature has been implemented.

NaturalSelect:
It has compatibility with older versions of code now.

NaturalSelect:
I explain my compatibility scheme here.

We splits the value of `maxClientIDKey` to `termPart` and `counterPart`, there are 3 possibilities:

* If `termPart` < `current term`, we return simply.
* If `termPart` == `current term`, we set the initial value of  `clientCounterPart` to  `counterPart + 1`, and `counterPart` is max value of `uint32`, we set `clientCounterConflict` flag to true.
* If `termPart` > `current term`, we set `clientCounterConflict` flag to true.

leonrayang:
@NaturalSelect The idea of using term is still good, but we need to consider the issues of wraparound and forward compatibility, which make the problem a bit complicated. Let's reduce the pressure on Raft by using the method of batch adding IDs

NaturalSelect:
> @NaturalSelect The idea of using term is still good, but we need to consider the issues of wraparound and forward compatibility, which make the problem a bit complicated. Let's reduce the pressure on Raft by using the method of batch adding IDs

Ok, I will switch to batch adding scheme later.

NaturalSelect:
> @NaturalSelect The idea of using term is still good, but we need to consider the issues of wraparound and forward compatibility, which make the problem a bit complicated. Let's reduce the pressure on Raft by using the method of batch adding IDs

Hello @leonrayang.

I have switch to batch adding scheme now.

Waiting for your review.

Victor1319 (dismissed):
LGTM

NaturalSelect:
Hello @leonrayang and @Victor1319.

This pull request is ready for merge.

Victor1319 (approved):
LGTM

leonrayang (approved):
LGTM

PR #2122, opened 2023-06-09
fix(util): fix wrong action when parse config

### LINKED ISSUE #2372: [Bug]: cfg.GetInt() in `util` always return 0
**Describe the bug**

The `GetInt()` function in `util/config/config.go` always return 0:
```
// GetInt returns a int value for the config key.
func (c *Config) GetInt(key string) int {
	x, present := c.data[key]
	if !present {
		return 0
	}
	if result, isInt := x.(int); isInt {
		return result
	}
	return 0
}
```

### Affected configuration items:
master: `raftRecvBufSize`
```go
m.raftRecvBufSize = int(cfg.GetInt(cfgRaftRecvBufSize))
```

metanode: `raftRecvBufSize`, `nameResolveInterval`
```go
m.raftRecvBufSize = int(cfg.GetInt(cfgRaftRecvBufSize))
updateInterval := cfg.GetInt(configNameResolveInterval)
```

datanode: `raftRecvBufSize`, `nameResolveInterval`
```go
s.raftRecvBufSize = int(cfg.GetInt(CfgRaftRecvBufSize))
updateInterval := cfg.GetInt(configNameResolveInterval)
```

objectnode: `bStoreWriteThreads`, `bStoreReadThreads`
```go
wt := cfg.GetInt(ebsWriteThreads)
rt := cfg.GetInt(ebsReadThreads)
```

**Environment**
master, develop-3.3.0

**To Reproduce**
1. using master as an example, we can add a configuration item `raftRecvBufSize` in `master/api_service_test.go`
```go
func createDefaultMasterServerForTest() *Server {

	cfgJSON := `{
		"role": "master",
		"ip": "127.0.0.1",
		"listen": "8080",
		"prof":"10088",
		"id":"1",
		"peers": "1:127.0.0.1:8080",
		"retainLogs":"20000",
		"tickInterval":500,
		"electionTick":6,
		"logDir": "/tmp/cubefs/Logs",
		"logLevel":"DEBUG",
		"walDir":"/tmp/cubefs/raft",
		"storeDir":"/tmp/cubefs/rocksdbstore",
		"clusterName":"cubefs",
                "raftRecvBufSize": 4096 // add a configuration item
	}`

        // omit the rest of the code
}
```
2. run any test function that can trigger the function
3. received unexpected results
`master/server.go`
```go
m.raftRecvBufSize = int(cfg.GetInt(cfgRaftRecvBufSize))
fmt.Printf("raftRecvBufSize[%v]\n", m.raftRecvBufSize) // output: raftRecvBufSize[0]
```

**Expected behavior**
output: should be `raftRecvBufSize[4096]`

setcy:

**background:** json.Unmarshal converts json literal integer to float64 instead of int64 when taget type is of type interface{}, so the function below are always return 0.
```go
func (c *Config) GetInt(key string) int {
	x, present := c.data[key] // the type of 'x' is float64
	if !present {
		return 0
	}
	if result, isInt := x.(int); isInt { //isInt is always false, so the result will never be returned
		return result
	}
	return 0
}
```

So we can use the UseNumber method on json.Decoder, which causes all numbers to unmarshal as `json.Number`

Before fixing this bug, value of `m.raftRecvBufSize` will be always '0' 

**Which issue this PR fixes** 
fixes #2372

### DIFF
--- a/util/config/config.go
+++ b/util/config/config.go
@@ -21,7 +21,7 @@ import (
 	"log"
 	"os"
 	"path"
-	"strconv"
+	"strings"
 )
 
 const (
@@ -55,7 +55,9 @@ func LoadConfigFile(filename string) (*Config, error) {
 // LoadConfigString loads config information from a JSON string.
 func LoadConfigString(s string) *Config {
 	result := newConfig()
-	err := json.Unmarshal([]byte(s), &result.data)
+	decoder := json.NewDecoder(strings.NewReader(s))
+	decoder.UseNumber()
+	err := decoder.Decode(&result.data)
 	if err != nil {
 		log.Fatalf("error parsing config string %s: %s", s, err)
 	}
@@ -66,7 +68,9 @@ func (c *Config) parse(fileName string) error {
 	jsonFileBytes, err := ioutil.ReadFile(fileName)
 	c.Raw = jsonFileBytes
 	if err == nil {
-		err = json.Unmarshal(jsonFileBytes, &c.data)
+		decoder := json.NewDecoder(strings.NewReader(string(jsonFileBytes)))
+		decoder.UseNumber()
+		err = decoder.Decode(&c.data)
 	}
 	return err
 }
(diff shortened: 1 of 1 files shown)


true1064 (approved):

true1064:
LGTM

leonrayang (approved):

PR #2124, opened 2023-06-10
[Feature]: Master does snapshot optimization

### LINKED ISSUE #1916: [Feature]: Master does snapshot optimization
- Consider cleaning up the rocksdb solution and make some optimizations, such as dropping the entire rocksdb and reinitializing.

**Additional context**
The current master's follower needs to clean up the rocksdb before accepting the leader's snapshot to prevent the information that has expired on the leader from being retained.

NaturalSelect:

* Optimize raft snapshot in master.
* Expose the rocksdb `close` interface.

The Benchmark Code:
```go
const volForSnapshot = "snapshtVol"

const volForSnapshotCount = 300

func BenchmarkSnapshot(b *testing.B) {
	var err error
	// perpare status
	req := &createVolReq{
		name:             "",
		owner:            "cfs",
		size:             1,
		mpCount:          1,
		dpReplicaNum:     3,
		capacity:         300,
		followerRead:     false,
		authenticate:     false,
		crossZone:        false,
		normalZonesFirst: false,
		zoneName:         testZone2,
		description:      "",
		qosLimitArgs:     &qosArgs{},
	}
	for i := 0; i != volForSnapshotCount; i++ {
		req.name = fmt.Sprintf("%v_%v", volForSnapshot, i)
		server.cluster.createVol(req)
	}
	time.Sleep(5 * time.Second)
	mdSnapshot, err := server.cluster.fsm.Snapshot()
	if err != nil {
		b.Error(err)
		return
	}
	b.Logf("snapshot apply index[%v]\n", mdSnapshot.ApplyIndex())
	s := &Server{}

	var dbStore *raftstore.RocksDBStore
	dbStore, err = raftstore.NewRocksDBStore("/tmp/cubefs/raft3", LRUCacheSize, WriteBufferSize)
	if err != nil {
		b.Fatalf("init rocks db store fail cause: %v", err)
	}
	defer dbStore.Close()
	fsm := &MetadataFsm{
		rs:    server.fsm.rs,
		store: dbStore,
	}
	fsm.registerApplySnapshotHandler(func() {
		fsm.restore()
	})
	s.fsm = fsm
	peers := make([]rproto.Peer, 0, len(server.config.peers))
	for _, peer := range server.config.peers {
		peers = append(peers, peer.Peer)
	}
	b.StopTimer()
	b.ResetTimer()
	b.StartTimer()
	if err = fsm.ApplySnapshot(peers, mdSnapshot); err != nil {
		b.Error(err)
		return
	}
	b.StopTimer()
	if fsm.applied != mdSnapshot.ApplyIndex() {
		b.Errorf("applied not equal,applied[%v],snapshot applied[%v]\n", fsm.applied, mdSnapshot.ApplyIndex())
		return
	}
	mdSnapshot.Close()
}
```

**Before**:
```log
1000000000	         0.2425 ns/op	       0 B/op	       0 allocs/op
--- BENCH: BenchmarkSnapshot-12
    master_manager_test.go:97: snapshot apply index[17517]
    master_manager_test.go:97: snapshot apply index[17517]
    master_manager_test.go:97: snapshot apply index[17517]
    master_manager_test.go:97: snapshot apply index[17517]
    master_manager_test.go:97: snapshot apply index[17517]
    master_manager_test.go:97: snapshot apply index[17517]
    master_manager_test.go:97: snapshot apply index[17517]
    master_manager_test.go:97: snapshot apply index[17517]
    master_manager_test.go:97: snapshot apply index[17517]
    master_manager_test.go:97: snapshot apply index[17517]
	... [output truncated]
PASS
```

The full log could be found in [here](https://github.com/NaturalSelect/logs/blob/main/bench_after.log).

**After:**
```log
1000000000	         0.1297 ns/op	       0 B/op	       0 allocs/op
--- BENCH: BenchmarkSnapshot-12
    master_manager_test.go:97: snapshot apply index[17517]
    master_manager_test.go:97: snapshot apply index[17517]
    master_manager_test.go:97: snapshot apply index[17517]
    master_manager_test.go:97: snapshot apply index[17517]
    master_manager_test.go:97: snapshot apply index[17517]
    master_manager_test.go:97: snapshot apply index[17517]
    master_manager_test.go:97: snapshot apply index[17517]
    master_manager_test.go:97: snapshot apply index[17517]
    master_manager_test.go:97: snapshot apply index[17517]
    master_manager_test.go:97: snapshot apply index[17517]
	... [output truncated]
PASS
```
The full log could be found in [here](https://github.com/NaturalSelect/logs/blob/main/bench_after.log).

fixes #1916 

* Please contract me if this pull request have passed your review, then I will rebase and rearrange commit points.
* This scheme has bad performance when the status is small.

**Release note**:

```release-note
```

### DIFF
--- a/master/master_manager_test.go
+++ b/master/master_manager_test.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"testing"
+	"time"
 
 	rproto "github.com/cubefs/cubefs/depends/tiglabs/raft/proto"
 	"github.com/cubefs/cubefs/proto"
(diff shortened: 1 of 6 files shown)


leonrayang (approved):
LGTM

wuchunhuan (approved):
The way using tmpdb to optimize applysnapshot is what we about to do, but as you mentioned, "This scheme has bad performance when the status is small", so might as well keep the old way of deleting entries in this situation, and let a threshold to dicide which way to go? On the other hand, real product enviroment would have lots of entries. 
So looks good to me.

NaturalSelect:
> The way using tmpdb to optimize applysnapshot is what we about to do, but as you mentioned, "This scheme has bad performance when the status is small", so might as well keep the old way of deleting entries in this situation, and let a threshold to dicide which way to go? On the other hand, real product enviroment would have lots of entries. So looks good to me.

Hello @wuchunhuan.

Although it has bad performance in a small status, but the time of execution is acceptable.

Maybe I didn't explain the small status correctly.

Small status means that the amount of data originally stored in is small.

And if we count the count of rocksdb entires, we may cause all data to be loaded into memory.

So I may not agree with the scheme of counting the number of items first.

NaturalSelect:
Please don't merge.

I will rebase and rearrange commit points later.

NaturalSelect:
Hello @leonrayang and @wuchunhuan.

This pull request is ready for merge now.

Issue #2133, opened 2023-06-12
[Bug]: PutConnect may casue stream tokens lost in some case.


Victor1319:
**Describe the bug**
```
	if forceClose {
		err := stream.Close()
		if err == io.ErrClosedPipe {
			return
		}

		pool.MarkClosed(stream)
		return
	}
```
like the code above,  `stream.close()` maybe return `errClosedPipe` err in the cases, eg: the used session is already been broken, closed by the others, stream already been closed, and so on, once these case occur, `pool.markClosed` will not execute, which will cause token lost.

PR #2135, opened 2023-06-12
Add some optimized code

### LINKED ISSUE #2134: not use stream conn pool when delete extents in metanode.
when use connection in pool to delete different dp's extent, will cause too many idel goroutines and connect in datanode. And metanode don't need to use connections in pool, it already use smux to delete extent request, which is enough for performance.

not use connet pool when do dlete extent request.

Victor1319:
1. not use connect pool when delete extent in metanode.
2. fix smux connect close check bug.

 fixes #2134, #2133


wuchunhuan (approved):
LGTM

PR #2147, opened 2023-06-14
fix(meta&data): reduce tcp connections create op when many getConn hap…

### LINKED ISSUE #2146: Many smuxPool.getConnect happen at the same time, whch may cause a lot of connections.
**Describe the bug**
When many concurrent requests try to obtain smux connections at the same time, that will result in the creation of many tcp connections.

**To Reproduce**
use a testcase to get connections at the same time with 1000 goroutines.

**Expected behavior**
use little tcp connections as limited to process sudden surge of requests.

Victor1319:
reduce tcp connections creat op when many getConn happen at the same time.

fixes #2146

### DIFF
--- a/util/smux_conn_pool.go
+++ b/util/smux_conn_pool.go
@@ -391,7 +391,7 @@ func (p *SmuxPool) callCreate() (createCall *createSessCall) {
 		} else {
 			return
 		}
-	default:
+		//default:
 	}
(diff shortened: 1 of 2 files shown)


leonrayang (approved):
LGTM

baijiaruo (approved):

Issue #2171, opened 2023-06-16
[Bug]: Fix 2 quota related issues


baijiaruo:
**Describe the bug**
1.No error reported when creating quota using multiple directories with the same name
2.Deleting quota without revoking it first will cause directories to be undeletable

PR #2172, opened 2023-06-16
fix(cli): Fix 2 quota related issues

baijiaruo:
close: #2171

fix：1.No error reported when creating quota using multiple directories with the same name
2.Deleting quota without revoking it first will cause directories to be undeletable


Victor1319 (approved):
LGTM

leonrayang (approved):
LGTM

PR #2178, opened 2023-06-19
feat(metanode): optimize mp leader send snapshot to follower

true1064:
(1) add a configuration item for raft retainLogs, for metaNode; 
(2) add a configuration item to set which format version to use when mp raft leader send snapshot to follower; 
(3) when follower apply snapshot, won't fail if format snapshot version not match, just handle the part of fields it can recognize;

(2 )and (3) are the  focus of this revision，to handle the compatibility of snapshot format of meta partitifion when upgrading. 

### DIFF
--- a/metanode/metanode.go
+++ b/metanode/metanode.go
@@ -52,24 +52,26 @@ var (
 // The MetaNode manages the dentry and inode information of the meta partitions on a meta node.
 // The data consistency is ensured by Raft.
 type MetaNode struct {
-	nodeId            uint64
-	listen            string
-	bindIp            bool
-	metadataDir       string // root dir of the metaNode
-	raftDir           string // root dir of the raftStore log
-	metadataManager   MetadataManager
-	localAddr         string
-	clusterId         string
-	raftStore         raftstore.RaftStore
-	raftHeartbeatPort string
-	raftReplicatePort string
-	zoneName          string
-	httpStopC         chan uint8
-	smuxStopC         chan uint8
-	metrics           *MetaNodeMetrics
-	tickInterval      int
-	raftRecvBufSize   int
-	connectionCnt     int64
+	nodeId                    uint64
+	listen                    string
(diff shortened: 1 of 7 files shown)


metanode/metanode.go
```diff
@@ -248,6 +250,34 @@ func (m *MetaNode) parseConfig(cfg *config.Config) (err error) {
...
+			log.LogInfof("invalid config raftSyncSnapFormatVersion, using default[%v]", m.raftSyncSnapFormatVersion)
+		} else {
+			m.raftSyncSnapFormatVersion = raftSyncSnapFormatVersion
+			log.LogInfof("by config raftSyncSnapFormatVersion:[%v]", m.raftSyncSnapFormatVersion)
+		}
+	} else {
+		m.raftSyncSnapFormatVersion = SnapFormatVersion_1
+		log.LogInfof("using default raftSyncSnapFormatVersion[%v]", m.raftSyncSnapFormatVersion)
+	}
+	syslog.Println("conf raftSyncSnapFormatVersion=", m.raftSyncSnapFormatVersion)
```
xiangcai1215, line 278: 这个信息需要打印出来吗，还是只有日志就好了

util/config/config.go
```diff
@@ -157,7 +157,12 @@ func (c *Config) GetInt64(key string) int64 {
 	return 0
 }
 
-// GetBool returns a int64 value for the config key.
+func (c *Config) HasKey(key string) bool {
+	_, present := c.data[key]
```
xiangcai1215, line 161: present直接用ok呗，golang更常见一点

xiangcai1215 (approved):
两个评论看看要不要改，流程没问题

Victor1319 (approved):
LGTM

util/config/config.go
true1064: util/config/config.go这个文件里用的都是present，不改了

PR #2179, opened 2023-06-20
transaction need consider repeat commit or else return error while network not stable between client and server

leonrayang:

transaction need consider repeat commit or else return error while network not stable between client and server


baijiaruo (approved):

Victor1319 (approved):
LGTM

PR #2181, opened 2023-06-20
set dp status readwrite when load from disk.

Victor1319:
1. set dp status readwrite
2. not update datapartition cache before load meta from disk ready.

**Release note**:

```release-note
```


leonrayang (approved):
LGTM

PR #2185, opened 2023-06-21
fix(metanode_tx): deadlock happend while process expired transactions

leonrayang:
deadlock happend while process expired transactions


true1064:
LGTM

true1064 (approved):

PR #2186, opened 2023-06-21
bugfix(client):serverRequest always return no vailed master if http s…

bboyCH4:
…tatus code is not 403 or 200
function serverRequest of master client sdk always return no vailed master if http status code is not 403 or 200


leonrayang (approved):

leonrayang:
LGTM

PR #2187, opened 2023-06-25
fix bug and add test to transaction expirition check process

leonrayang:

fix bug and add test to transaction expirition check process
1)fix(transcation):tx expire check goroutine may start multi times on parrallel
2)test(metanode_transaction): processExpiredTransactions may run mulit times on parrallel if leader change frequncey
3)feat(transaction):enable pause transaction including the expire process process in metanode


baijiaruo (approved):

true1064 (approved):

true1064:
LGTM

Issue #2189, opened 2023-06-25
[Bug]: Fixed some quota related issues


baijiaruo:
**Describe the bug**
1.Subdirectories under the quota directory cannot be moved
2.After creating the quota, delete the quota root directory and create it again, but the apply still succeeds
3.Multi-level directory setting quota report does not has only one /
4.2000w directory file after the quota is created apply error

PR #2190, opened 2023-06-25
fix(master/metanode/cli): Fix the following bugs

baijiaruo:
close: #2189

1.Subdirectories under the quota directory cannot be moved 2.After creating the quota, delete the quota root directory and create it again, but the apply still succeeds 3.Multi-level directory setting quota report does not has only one / 4.2000w directory file after the quota is created apply error


leonrayang (approved):
LGTM

Victor1319 (approved):
LGTMLGTM

PR #2196, opened 2023-06-26
feat(metanode):  meta partition raft leader send snapshot includes txInfo, txRbInode, txRbDentry only when snapshot format version is v1

true1064:
 optimization: meta partition raft leader send snapshot includes txInfo, txRbInode, txRbDentry only when snapshot format version is v1.


Victor1319 (approved):
LGTMLGTM

xiangcai1215 (approved):

PR #2198, opened 2023-06-27
fix(client.transaction):retry logic have no times limitation while get tx conflict from mp

leonrayang:
retry logic have no times limitation while get tx conflict from mp


true1064 (approved):

Victor1319 (approved):
LGTM

PR #2201, opened 2023-06-27
bugfix(master):if maxInode among the mp replicas become the same, rem…

bboyCH4:
…ove the entry from maxInodeNotEqualMP cache

If maxInode among the mp replicas become the same,  remove the entry from maxInodeNotEqualMP cache.
Then the cli tool would nerver display this entry


leonrayang (approved):
LGTM

PR #2221, opened 2023-06-30
Develop 3.3.0

true1064:
while meta node loading all the meta partitions, the program will abort if panic cccurs when loading any one all the partitions.
but we hope that meta node not to abort in such failure, it can just skip  partitions which failed to load and continue to load and serve other partitions.


Victor1319 (approved):
LGTM

leonrayang (approved):
LGTM

Issue #1580, opened 2022-10-27
[Bug Report] need add lock for snapshot


sjp00556:
The snapshot process and fsm apply are executed in parallel, if fsm apply is not idempotent such as fsmCreateDentry may cause wrong result.  we need add lock for snapshot

https://github.com/cubefs/cubefs/blob/7976d6b660624b74f8b0e757274ccfab11fe66d5/metanode/partition_item.go#L143

leonrayang:
sorry,I didn't got the point.The inner in make snapshot is based on btree clone,the modification of fsm routines have no influnce to snapshot. and the fsm apply must be idempotent.

sjp00556:
这里是主节点向从节点发送的snapshot，不是store_ticker定时5分钟raftfsm里打的snapshot。raft在向从节点sendAppend时如果节点落后太多时，调用 func (mp *metaPartition) Snapshot() 实时打的快照。 这里不再apply流程里，apply有个channel，是异步执行的，每个apply之间是串行的，但是和raft run Snapshot这里会并行 


![image](https://user-images.githubusercontent.com/6017354/198522782-4995c1d9-e108-4ebb-820c-0c319a42f089.png)

sjp00556:
![image](https://user-images.githubusercontent.com/6017354/198523573-e175ee87-ba5f-4e71-9640-b302bbf2748d.png)

比如这里newMetaItemIterator 的同时 Apply 在进行index=123的 fsmCreateDentry 操作。如果inodeTree clone 之后 CreateDentry 又创建了dentry，再mp.dentryTree.GetTree ，这时快照里的数据就包含了index=123未提交的部分修改。 但是fsmCreateDentry不是幂等的，就会导致从节点再次执行123的时候，dentry存在，但是parent inode Nlink没有加一

leonrayang:
I see， it's also wrong if the parent inode link may be added more than once, do you have any idea to fix it，please share it.

sjp00556:
> I see， it's also wrong if the parent inode link may be added more than once, do you have any idea to fix it，please share it.

simple way: add mutex lock for apply() and newMetaItemIterator
anther solution: send last local snapshot which created by store_ticker in fsm apply, but the cost is more disk read and log lag. we also need to retain more raft log to help slaver follow up the log lag

leonrayang:
welcome to  join us and make a pr.

gueFDF:
Does this problem still exist？, I think I can solve .

[...]

PR #2222, opened 2023-06-30
fix: dentry fsm create between inodeTree clone and deteyTree clone le…

leonrayang:
…ad to inconstent of link of parentInode

close:#1580. use lock to protect dentry create and snapshot clone parrellel

issue happend while create a file, and do fsm operation and clone as the sequece : 1)inode create 2)inodeTree clone  3)fsmCreateDetnry 4)dentryTree lone, while restart, the 3)step will not work because the detry already the detreeTree, which make the parent miss the link addition,so we need make the detry create not happend between the tree clone

**Release note**:

```release-note
```

### DIFF
--- a/build/build.sh
+++ b/build/build.sh
@@ -243,7 +243,6 @@ run_test() {
     pushd $SrcPath >/dev/null
     echo -n "${TPATH}"
 #    go test $MODFLAGS -ldflags "${LDFlags}" -cover ./master
-
     go test -cover -v -coverprofile=cover.output $(go list ./... | grep -v depends | grep -v master) | tee cubefs_unittest.output
     ret=$?
     popd >/dev/null
--- a/metanode/partition.go
+++ b/metanode/partition.go
@@ -456,6 +456,7 @@ type metaPartition struct {
 	uidManager             *UidManager
 	xattrLock              sync.Mutex
 	mqMgr                  *MetaQuotaManager
+	nonIdempotent          sync.RWMutex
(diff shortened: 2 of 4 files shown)


build/build.sh
```diff
@@ -243,8 +243,8 @@ run_test() {
     pushd $SrcPath >/dev/null
     echo -n "${TPATH}"
 #    go test $MODFLAGS -ldflags "${LDFlags}" -cover ./master
-
-    go test -cover -v -coverprofile=cover.output $(go list ./... | grep -v depends | grep -v master) | tee cubefs_unittest.output
+go test -run TestMpSnapshotLock -cover -v -coverprofile=cover.output $(go list ./... |grep meta) | tee cubefs_unittest.output
```
Victor1319, line 246: this is your local tmp code?

metanode/partition.go
```diff
@@ -456,6 +456,20 @@ type metaPartition struct {
...
+
+func (mp *metaPartition) nonIdemPtLock() {
+	for {
+		if atomic.CompareAndSwapInt32(&mp.nonIdempotent, 0, 1) {
+			break
+		}
+		time.Sleep(time.Millisecond * 10)
+	}
+}
+
```
Victor1319, line 470: sugget use sync.mutex directly

Victor1319 (approved):
LGTM

xiangcai1215 (approved):

PR #2232, opened 2023-07-03
fix(raft): when raft replicas or raft partitions are deleted, delete them from raft monitor accordingly

true1064:
…them from raft monitor accordingly

the bug: 
when raft replicas or raft partitions are deleted, them are still in raft monitor and the error log is still printing.

fix:
when raft replicas or raft partitions are deleted, delete them from raft monitor accordingly


Victor1319 (approved):
LGTM

xiangcai1215 (approved):

Issue #2242, opened 2023-07-04
[Bug]: Data content read by bcache is incorrect during AI training


bboyCH4:
**Describe the bug**
During AI high-concurrency training tasks, the data content read by bcache is incorrect, resulting in the interruption of the training task.

**Environment**
3.2.1

true1064:
done: https://github.com/cubefs/cubefs/pull/2243

PR #2243, opened 2023-07-05
bugfix(client):check bcache cache file path

bboyCH4:
close:#2242

Client would check cache path received from bcache service, in case   cache path belongs to other file


leonrayang (approved):
LGTM

true1064 (approved):

true1064:
LGTM

Issue #2255, opened 2023-07-07
[Feature]: The original single directory limit is changed to dirLimit instead of dirQuota


baijiaruo:
**Is your feature request related to a problem? Please describe.**
Because there is already a feature of directory quotas, and there may be ambiguity in code understanding with the original limit on the number of single directories, so optimize

PR #2261, opened 2023-07-07
feat(master): The original single directory limit is changed to dirLimit instead of dirQuota

baijiaruo:
close: #2255

The original single directory limit is changed to dirLimit instead of dirQuota


Victor1319 (approved):
LGTM

Victor1319 (approved):
LGTM

leonrayang (approved):
LGTM

Issue #2258, opened 2023-07-07
[Bug]: In libsdk.go, freeCache for writer buffer is not called when file is closed, which causes oom.


M1eyu2018:
**Describe the bug**
In libsdk.go, freeCache for writer buffer is not called when file is closed, which causes oom.

Issue #2257, opened 2023-07-07
[Bug]:  In Fuse client, a buffer will be allocated to two writer of different files if this buffer was freed twice by last writer, which causes inconsistence.


M1eyu2018:
**Describe the bug**
In Fuse client, a buffer will be allocated to two writer of different files if this buffer was freed twice by last writer, which causes inconsistence.

true1064:
Done: https://github.com/cubefs/cubefs/pull/2265

PR #2262, opened 2023-07-10
bugfix(client):fix the logic of returning the cache to the cache pool. 

bboyCH4:
close: #2258 #2257

1.In cold volume scenario, fwriter is not released when file is opened in write mode first and then opened in read-only mode.So when file is closed, fwriter's buf is returned to cache pool twice. This will result in subsequent write operations obtaining two identical buffer addresses from the cache pool. 2. closeStream function will return fileWriter's buf to cache pool to avoid OOM.

### DIFF
--- a/client/fs/file.go
+++ b/client/fs/file.go
@@ -98,7 +98,6 @@ func NewFile(s *Super, i *proto.InodeInfo, flag uint32, pino uint64, filename st
 			fReader = blobstore.NewReader(clientConf)
 		case syscall.O_WRONLY:
 			fWriter = blobstore.NewWriter(clientConf)
-
 		case syscall.O_RDWR:
 			fReader = blobstore.NewReader(clientConf)
 			fWriter = blobstore.NewWriter(clientConf)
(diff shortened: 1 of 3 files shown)


client/fs/file.go
```diff
@@ -96,9 +96,10 @@ func NewFile(s *Super, i *proto.InodeInfo, flag uint32, pino uint64, filename st
 		switch flag {
 		case syscall.O_RDONLY:
 			fReader = blobstore.NewReader(clientConf)
+			fWriter = nil
```
sejust, line 99: Redundant code.

client/fs/file.go
```diff
@@ -96,9 +96,10 @@ func NewFile(s *Super, i *proto.InodeInfo, flag uint32, pino uint64, filename st
...
 		case syscall.O_WRONLY:
 			fWriter = blobstore.NewWriter(clientConf)
-
+			fReader = nil
```
sejust, line 102: Redundant code.

client/fs/file.go
```diff
@@ -260,8 +261,10 @@ func (f *File) Open(ctx context.Context, req *fuse.OpenRequest, resp *fuse.OpenR
 		switch req.Flags & 0x0f {
 		case syscall.O_RDONLY:
 			f.fReader = blobstore.NewReader(clientConf)
+			f.fWriter = nil
```
sejust, line 264: MUST Close it before reset to nil.

client/fs/file.go
```diff
@@ -260,8 +261,10 @@ func (f *File) Open(ctx context.Context, req *fuse.OpenRequest, resp *fuse.OpenR
...
 		case syscall.O_WRONLY:
 			f.fWriter = blobstore.NewWriter(clientConf)
+			f.fReader = nil
```
sejust, line 267: Close it before reset to nil.

client/fs/file.go
```diff
@@ -283,7 +286,9 @@ func (f *File) Release(ctx context.Context, req *fuse.ReleaseRequest) (err error
 
 	defer func() {
 		stat.EndStat("Release", err, bgTime, 1)
-		f.fWriter.FreeCache()
+		if f.fWriter != nil {
+			f.fWriter.FreeCache()
```
sejust, line 290: Check `Write` is nil or not in its function `FreeCache`.

zhangtianjiong (dismissed):
LGTM

sdk/data/blobstore/writer.go
```diff
@@ -619,7 +619,10 @@ func (writer *Writer) FreeCache() {
 	if buf.CachePool == nil {
 		return
 	}
-	buf.CachePool.Put(writer.buf)
+	writer.once.Do(func() {
+		buf.CachePool.Put(writer.buf)
+		writer.buf = nil
```
sejust, line 624: Better:
```
buf := writer.buf
writer.buf = nil
buf.CachePool.Put(buf)
```

sejust (changes requested):

zhangtianjiong (approved):

zhangtianjiong (dismissed):

zhangtianjiong (approved):

PR #2263, opened 2023-07-10
Log: Dir of log(normal|statistic|audit) be grant as 755 for other use…

leonrayang:
…r access

For other users in linux, the cubefs logs should be readable, which facilitates the division of operational levels.


true1064 (approved):

true1064:
LGTM

PR #2264, opened 2023-07-10
fix alloc volume failed but proxy has available volumes

JasonHu520:
fix alloc volume failed but proxy has available volumes


blobstore/access/controller/service_test.go
```diff
@@ -83,7 +83,7 @@ func TestAccessServiceGetServiceHost(t *testing.T) {
 	require.ElementsMatch(t, []string{"proxy-1", "proxy-2"}, hosts)
 }
 
-func TestAccessServicePunishService(t *testing.T) {
+func TestAccessServicePunishServicex(t *testing.T) {
```
sejust, line 86: I was changed this on `blobstore-v1.2.3`, please update it.

sejust (changes requested):

sejust (approved):

PR #2265, opened 2023-07-10
bugfix(client): Fix the logic of returning the cache to the cache pool

bboyCH4:
1.In cold volume scenario, fwriter is not released when file is opened in write mode first and then opened in read-only mode.So when file is closed, fwriter's buf is returned to cache pool twice. This will result in subsequent write operations obtaining two identical buffer addresses from the cache pool. 2.CloseStream function will return fileWriter's buf to cache pool to avoid OOM.

close: #2258 #2257


zhangtianjiong (approved):

leonrayang (approved):
LGTM

PR #2275, opened 2023-07-12
fix(sdk): fix uploading through s3 does not immediately count the quota

baijiaruo:
close: 2274

fix uploading through s3 does not immediately count the quota


tangdeyi (approved):
LGTM

Victor1319 (approved):
LGTM

PR #2279, opened 2023-07-12
test(cli): support unit test for cli

setcy:
## **What this PR does / why we need it**:
Support unit test for cli.

## How to mock data
I added many interfaces in SDK, and cli only calls these interfaces instead of directly calling the original methods. By this way, we can easily mock the data in the cli module. The data used for mocking is written in `cli/mocktest` folder.

```go
type IMasterClient interface {
	AddNode(address string)
	Leader() (addr string)
	AdminAPI() IAdminAPI
	ClientAPI() IClientAPI
	NodeAPI() INodeAPI
	UserAPI() IUserAPI
	SetLeader(addr string)
	SetTimeout(timeout uint16)
	Nodes() (nodes []string)
}

type IClientAPI interface {
	GetVolume(volName string, authKey string) (vv *proto.VolView, err error)
	GetVolumeWithoutAuthKey(volName string) (vv *proto.VolView, err error)
	GetVolumeWithAuthnode(volName string, authKey string, token string, decoder Decoder) (vv *proto.VolView, err error)
	GetVolumeStat(volName string) (info *proto.VolStatInfo, err error)
	GetMetaPartition(partitionID uint64) (partition *proto.MetaPartitionInfo, err error)
	GetMetaPartitions(volName string) (views []*proto.MetaPartitionView, err error)
	GetDataPartitions(volName string) (view *proto.DataPartitionsView, err error)
	GetPreLoadDataPartitions(volName string) (view *proto.DataPartitionsView, err error)
}
```

## How to record errors

I rewrited the `errout` function to support modifying the handling of errors. By default, when the `errout` function is called, it will automatically execute the `OsExitWithLogFlush` function, and the program will stop immediately. When test files start running, we will change the `erroutHandler` so that program won't stop and the error will be recorded.

```go
var erroutHandler = OsExitWithLogFlush

func errout(format string, a ...interface{}) {
	log.LogErrorf(format, a...)
	_, _ = fmt.Fprintf(os.Stderr, format, a...)
	erroutHandler(format, a)
}

func OsExitWithLogFlush(_ string, _ ...interface{}) {
	log.LogFlush()
	os.Exit(1)
}
```
```go
func setupTestErrorRecorder() *testErrorRecorder {
	recorder := &testErrorRecorder{}
	erroutHandler = func(format string, args ...interface{}) {
		_, _ = fmt.Fprintf(recorder, format, args...)
	}
	return recorder
}
```

## How to write test cases

I defined a TestCase for unified management of test case execution. In most of the test functions, we only need to write the test case slice and then call the `runTestCases` function for execution.

```go
type TestCase struct {
	name      string    // name of test case
	args      []string  // arguments of test command
	expectErr bool      // whether the error is expected
}

func runTestCases(t *testing.T, testCases []*TestCase) {
	for _, tc := range testCases {
		t.Run(tc.name, func(t *testing.T) {
			err := testRun(tc.args...)
			if tc.expectErr {
				assert.Error(t, err)
			} else {
				assert.NoError(t, err)
			}
		})
	}
}
```
```go
func TestClusterInfoCmd(t *testing.T) {
	testCases := []*TestCase{
		{
			name:      "Valid arguments",
			args:      []string{"cluster", "info"},
			expectErr: false,
		},
	}

	runTestCases(t, testCases)
}
```


xiangcai1215 (approved):
LGTM

setcy:
ready for review

PR #2283, opened 2023-07-12
fix(metanode): deduplicate link and unlink requests

### LINKED ISSUE #2256: [Bug]: metanode:  mv files by rename may cause inode lose when meta timeout retry
**Describe the bug**
without transaction, ilink() and iunlink() is not idempotent. when sdk timeout, rpc retry may cause link nums wrong. Especially in mv file may cause inode lose. 

rename_ll will first use ilink() to increase links and then do rename options and finaly use iunlink() to decrease links.  if iunlink call twice (first call is failed, maybe caused by metanode restart or disk slow), this inode's link num will be decreased to 0.

use transaction is too heavy, so  we need deduplicate ilink() and iunlink() calls, make them idempotent

**To Reproduce**
add random sleep time in iunlink function, will reporduce this problem
ac6bf2f1-6de7-4bf9-9284-bc87cf175145

**Expected behavior**
need deduplicate ilink() and iunlink() calls, make them idempotent

sjp00556:
introduce request uniq id, use uniq request id to dedup client's rpc retries

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #2256


sjp00556:
**Test Plans:**

1. client send unlink/link rpc retries can be deduplicate, metanode's uniqChecker keeps uniqids at least 5 minutes, if uniqids num lease than 1024，they will be kept much longer which can cover slow disk situation.
image

2. partition snapshot correctly stored. after metanode restart, it can load max uniqid and load uniqchecker data to rebuild deduplication map and queue
image

3. send and apply snapshot correctly, when add partition replica or raft log fall behind too much.

leonrayang (dismissed):
LGTM

leonrayang (approved):
LGTM

Issue #2310, opened 2023-07-20
[Bug]: Update volume operation causes quota switch to be off


baijiaruo:
**Describe the bug**
When updating the volume operation and modifying other parameters of the volume, the quota will be set to false

PR #2311, opened 2023-07-20
fix(cli): update volume operation cause quota switch to be off

baijiaruo:
close: #2310
Signed-off-by: baijiaruo <baijiaruo@126.com>

When updating the volume operation and modifying other parameters of the volume, the quota will be set to false


Victor1319 (approved):
LGTM

true1064 (approved):

true1064:
LGTM

PR #2331, opened 2023-07-24
fix(metanode): fix metanode's compatibility after uniqId is introduced:

true1064:
fix metanode's compatibility after uniqId is introduced:
(1) when meta partition loading local snapshot files created by elder versions, skip uniqId related stuff and load will succeed;
(2) when meta partition raft leader send snapshot to followers, send uniqId related stuff only when uniqId is not 0; 
(3) when meta partition raft follower apply snapshot from the leader and  encounter unknown op code, not return err if leader's snap format version it bigger than follower's.

**Release note**:

```release-note
```


metanode/partition_store.go
```diff
@@ -670,9 +660,43 @@ func (mp *metaPartition) loadTxID(rootDir string) (err error) {
...
-	defer log.LogInfof("loadUniqChecker partition-%v complete", mp.config.PartitionId)
+func (mp *metaPartition) loadUniqID(rootDir string) (err error) {
+	filename := path.Join(rootDir, uniqIDFile)
+	if _, err = os.Stat(filename); err != nil {
+		err = nil
+		return
+	}
+	data, err := ioutil.ReadFile(filename)
+	if err != nil {
+		if err == os.ErrNotExist {
```
Victor1319, line 671: here, notExist error should also return err.

true1064: fixed

Victor1319 (approved):
LGTM

xiangcai1215 (approved):

Issue #2338, opened 2023-07-25
[Feature]: libsdk support quota


baijiaruo:
**Is your feature request related to a problem? Please describe.**
In some scenarios using libsdk need to support quota

PR #2339, opened 2023-07-25
feat(libsdk): libsdk support quota

baijiaruo:
close: #2338

In some scenarios using libsdk need to support quota


Victor1319 (approved):
LGTM

leonrayang (approved):
LGTM

Issue #2341, opened 2023-07-26
[Feature]: Inode deletion support lock time


shuqiang-zheng:
**Is your feature request related to a problem? Please describe.**
In some scenarios using Inode deletion need to support lock time

PR #2348, opened 2023-07-26
feat(Inode Lock Time): Added support for delete-lock-time to Inode.

shuqiang-zheng:
close: #2341

feat(Inode Lock Time): Added support for delete-lock-time to Inode.


leonrayang (approved):
LGTM

PR #2361, opened 2023-07-27
fix(meta): do retry in the case that submit to raft fail when raft no…

### LINKED ISSUE #2360: [Bug]: the error return is not right when submit to raft failed.
**Describe the bug**
The error return is not right when submit to raft failed, so client will not do retry request when no leader, and will fail.

**Environment**
develop-3.3.0

**To Reproduce**

**Expected behavior**

**Additional context**

Victor1319:
… leader.

Closes #2360

The error return is not right when submit to raft failed, so client will not do retry request when no leader, and will fail.

**Release note**:

```release-note
```


baijiaruo (approved):
LGTM

leonrayang (approved):
LGTM

PR #2373, opened 2023-07-29
fix(workflow): issue permission changes to `write`

setcy:
When github action executes these two files, an error will be reported:
```
The workflow is not valid. .github/workflows/issue_similarity.yml (Line: 8, Col: 11): Unexpected value 'read|write'
The workflow is not valid. .github/workflows/issue_inactive.yml (Line: 10, Col: 11): Unexpected value 'read|write'
```

`read|write` is not a valid permission setting, you need to specify `read` or `write` for each permission. We have to change it to the following code:
```
permissions:
   issues: write
```
This means that this GitHub Action has the right to write (and read) issues. The `write` permission will include the `read` permission

### DIFF
--- a/.github/workflows/issue_inactive.yml
+++ b/.github/workflows/issue_inactive.yml
@@ -7,7 +7,7 @@ on:
 # labels: inactive-one-week, inactive-one-month, inactive-three-month, inactive-one-year, inactive-two-year
 
 permissions:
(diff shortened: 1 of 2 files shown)


sejust (approved):

xiaochunhe (approved):
LGTM

Issue #2383, opened 2023-08-01
[Feature]: 


xiangcai1215:
**Is your feature request related to a problem? Please describe.**
Currently, CubeFS's Raft implementation does not have the ability to perform leader pre-elections. This may result in scenarios where some nodes constantly increase their current Term value in order to trigger leader elections during network isolation. After the network is restored, an unnecessary leader election process may be triggered.

**Describe the solution you'd like**
Adding a pre-candidate role and a pre-election process can help address the issue of unnecessary leader elections triggered by constantly increasing the current Term value during network isolation. In this proposed solution, a node must first become a pre-candidate and successfully complete the pre-election process before increasing its current Term value and triggering a leader election process.

During the pre-election process, a pre-candidate sends RequestVoteRPCs to all other nodes in the cluster. If the pre-candidate receives votes from a majority of the nodes, it becomes a candidate and begins a regular leader election process to become the new leader.

**Describe alternatives you've considered**

**Additional context**
By adding a pre-candidate role and pre-election process, CubeFS's Raft implementation can avoid unnecessary leader elections and improve the efficiency and reliability of the system.

true1064:
thumbs up

PR #2385, opened 2023-08-01
[Feature]: add Pre-vote capability to CubeFS Raft 

xiangcai1215:
CubeFS's MetaNode, DataNode, and Master components all use Raft. However, in exceptional situations such as network partitioning, the communication between nodes may be hindered, resulting in unnecessary leader election processes .

To address this issue, we use pre-Vote which involves conducting a preliminary round of voting before the formal election. Only when the candidate receives the support of the majority of nodes in the pre-vote can it participate in the formal election. This can reduce unnecessary leader election processes, improve system efficiency, and reliability.

add pre

fixes：#2383

**Release note**:

```release-note
```
When upgrading versions, it is important to ensure that the Raft service runs smoothly in a cluster composed of both old and new versions.

### DIFF
--- a/depends/tiglabs/raft/config.go
+++ b/depends/tiglabs/raft/config.go
@@ -89,7 +89,12 @@ type Config struct {
 	RetainLogs uint64
 	// LeaseCheck whether to use the lease mechanism.
 	// The default value is false.
+	// (this equal etcd's raft checkQuorum)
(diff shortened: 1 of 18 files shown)


JasonHu520 (approved):

leonrayang (approved):
LGTM

PR #2386, opened 2023-08-01
[Feature]: client support change the explicit name of the filesystem

setcy:

Client support change the explicit name of the filesystem.

The default is still `"cubefs-" + [volume name]`

## example: 

**configuration file:**
``` json
{
  "masterAddr": "172.16.1.101:17010,172.16.1.102:17010,172.16.1.103:17010",
  "mountPoint": "/home/data/client/mnt",
  "volName": "vol_cross5",
  "owner": "cfs",
  "logDir": "/home/data/client/log",
  "logLevel": "debug",
  "fileSystemName": "testfs"
}
```

**mount the volume:**
```bash
cfs-client -c client.json
```

**result:**
```bash
$ df -h
Filesystem      Size  Used Avail Use% Mounted on
testfs         1000G     0 1000G   0% /home/data/client/mnt
```
```bash
$ mount
testfs on /home/data/client/mnt type fuse.cubefs (rw,nosuid,nodev,relatime,user_id=0,group_id=0,allow_other)
```

### DIFF
--- a/client/fuse.go
+++ b/client/fuse.go
@@ -644,10 +644,10 @@ func mount(opt *proto.MountOptions) (fsConn *fuse.Conn, super *cfs.Super, err er
 		fuse.MaxReadahead(MaxReadAhead),
 		fuse.AsyncRead(),
 		fuse.AutoInvalData(opt.AutoInvalData),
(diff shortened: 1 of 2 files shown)


true1064 (approved):
LGTM

PR #2387, opened 2023-08-01
fix(shell): do not overwrite configuration file if file exists

setcy:

When I am trying [Install Basic Cluster](https://cubefs.io/docs/master/deploy/node.html#install-basic-cluster) in Standalone Mode, I discovered when I running this command:
```bash
# Start the script
sh ./shell/depoly.sh /home/data bond0
```
the configuration file in `/home/data/conf` will be overwrited.

So this pr is to ensure that if there is a configuration file in this directory, it will not continue to generate. In this way, we can easily test deployments **under different configuration files** by modifying these configuration files.

### DIFF
--- a/shell/depoly.sh
+++ b/shell/depoly.sh
@@ -1,4 +1,4 @@
-#/bin/bash
+#!/bin/bash
 
 if [ $# -ne 2 ]; then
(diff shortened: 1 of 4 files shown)


xiangcai1215 (approved):
LGTM

PR #2398, opened 2023-08-02
feat(metanode,sdk): Optimize quota performance

baijiaruo:
close: 2393

Quota interacts with metanode twice when creating and deleting files, and reduces one interaction by caching


leonrayang (approved):
LGTM

Issue #2399, opened 2023-08-02
[Bug]: Log can not cleared in a timely manner


shuqiang-zheng:
**Describe the bug**
The current log service may not be able to clean up the log files in a timely manner, in most cases when the log service is started，it will take the twenty percent of current disk available space as logLeftSpaceLimit, but when the current disk space  available is small and at this time the service happens to restart, the logLeftSpaceLimit now will bealso  very small, which may result in the existence of a large number of logs can not be deleted!
**Environment**
develop-3.3.0

**To Reproduce**

**Expected behavior**
The logging service should start with a minimum LogLeftSpaceLimit to ensure that log deletion can be triggered if the available space is less than this value.

**Additional context**

true1064:
already merged into branch master

PR #2394, opened 2023-08-02
fix(log): fix bug that logs were not cleaned in time.

shuqiang-zheng:
close: #2399
**What this PR does / why we need it**:
fix bug that logs were not cleaned in time.
**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #

**Special notes for your reviewer**:

**Release note**:

```release-note
```

PR #2408, opened 2023-08-03
fix(log): fix bug that logs were not cleaned in time.

shuqiang-zheng:
Because the available space on the disk where log directory is located is not certain, the original unit test may have caused the set logLeftSpaceLimit to fail.So there was a problem with the unit test in pr #2394, this pr fixes the problem with the unit test  so that the log's test cases can  pass the test for all available space scenarios.
fixes: #2399  

**Release note**:

```release-note
```

### DIFF
--- a/blockcache/cmd.go
+++ b/blockcache/cmd.go
@@ -179,7 +179,7 @@ func main() {
 		level = log.ErrorLevel
 	}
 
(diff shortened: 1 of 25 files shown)


Victor1319 (approved):
LGTM

PR #2414, opened 2023-08-04
support to clear orphan tx info.

### LINKED ISSUE #2410: [Bug]: In case of network congestion, there might be orphan tx in RM node.
**Describe the bug**
If an abnormal register tx request is recived by RM after the tx is finised, tow minutes laster, the tx info in RM will be orhpan item.

**Environment**
develop-3.3.0

**To Reproduce**

**Expected behavior**
there is a process to clear the clear orphan tx item.

**Additional context**

### LINKED ISSUE #2411: [Feature]: Support the control of automatic creation for new partitions.
In some case, the cluster will create lots of partitions becasue abnormal execption, and most of partitons is no use.
The system disableAutoCreate swicth once is open, can't create volume and create partitions.

Add a switch, once it is open, not support partitions automaticaly, but the users can create partitions or volume by manual.

Victor1319:
1. fix the bug that update volume tast exit because of request master error once.
2. add switch to control whether enable automatic create partition.
3. support clear orphan tx in rm caused by network problem.
4. update the return err msg to avoid return empty msg.
fixes #2411 
fixes #2410

**Release note**:

```release-note
```

### DIFF
--- a/client/fuse.go
+++ b/client/fuse.go
@@ -623,22 +623,19 @@ func mount(opt *proto.MountOptions) (fsConn *fuse.Conn, super *cfs.Super, err er
 		var mc = master.NewMasterClientFromString(opt.Master, false)
 		t := time.NewTicker(UpdateConfInterval)
 		defer t.Stop()
-		for {
-			select {
-			case <-t.C:
-				log.LogDebugf("UpdateVolConf: load conf from master")
-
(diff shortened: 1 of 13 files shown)


leonrayang (approved):
LGTM

baijiaruo (approved):
LGTM

PR #2420, opened 2023-08-04
fix(log): fix bug that log configuration of logLeftSpaceLimit parsing…

shuqiang-zheng:
fix bug that log configuration of logLeftSpaceLimit parsing failed, and debug log when log deletion is triggered.Because the getInt() method in the current log configuration file parsing can not correctly parse the configuration value, so it will lead to the configuration item can not take effect properly, so modified the parsing method of these configuration items, in addition, the lack of debug logs in the log deletion process, it is impossible to know whether the deletion process is triggered or not.
fixes: #2399 

**Release note**:

```release-note
```

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -25,6 +25,7 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 	"syscall"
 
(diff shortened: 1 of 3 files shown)


Victor1319 (approved):
LGTM

leonrayang (approved):
LGTM

PR #2423, opened 2023-08-07
fix retry unlink dir inode failed bug.

Victor1319:
**What this PR does / why we need it**:
check error fisrt to avoid panic when metanode request master. fixes #2421
when unlink dir ino, return ino info in rbInode. fixes #2422 
**Special notes for your reviewer**:

**Release note**:

```release-note
```

PR #2425, opened 2023-08-07
fix(meta)：check if err occur when query master, in function checkLocalPartitionMatchWithMaster() 

true1064:
in checkLocalPartitionMatchWithMaster()，if failed to query master may trigger panic for not checking if err occur.
not add checking if err occur before reading the response var.


true1064:
already fixed in branch develop-3.3.0 by another guy，PR: https://github.com/cubefs/cubefs/pull/2423

PR #2428, opened 2023-08-07
fix(Raft): fix the raft log append bug

xiangcai1215:

1. When follower's log confit with leader's log. The follower always returns the log's last index as rejected index which lead to leader alaways send the same log to follower, which means the logs will never be synchronized. 

1. Fixes the confit log error by setting rejected index to min(log.LastIndex(), m.index).
2. Adding a feature will make it easier to find the conflict index. Therefore, there is no need to reduce indexes one by one when leader replica logs to follower. 

fixes#2376

**Release note**:

```release-note
```


leonrayang (approved):
LGTM

yhjiango (approved):
LGTM

tangdeyi (approved):
LGTM

PR #2458, opened 2023-08-09
fix(meta): deleteMarkedInodes() will panic when print log if inode.mu…

true1064:
fix the problem that deleteMarkedInodes() will panic when print log if inode.multiSnap is nil


Victor1319 (dismissed):
LGTM

Victor1319 (approved):
LGTM

leonrayang (approved):
LGTM

PR #2461, opened 2023-08-09
fix(master): fix panic that may be caused in lcnode task

honeyvinnie:
fix panic that may be caused in lcnode task and optimized some code


leonrayang (approved):
LGTM

PR #2467, opened 2023-08-10
feat(Inode Lock Time): Changed the default value of optDeleteLockTime…

shuqiang-zheng:
close: #2341

Currently, if you don't set the delete lock time when updating a volume, the delete lock function will be turned off when updating other information because its default value is 0. So this commit changes the default value of optDeleteLockTime and the judgment criteria.What's more, this commit solved the problem of deletion lock failure after the client's system time has been changed.
fixes: #2341 

**Release note**:

```release-note
```


Victor1319 (approved):
LGTM

leonrayang (approved):
LGTM

PR #2483, opened 2023-08-14
fix(doc): the value of "profPort" in client's configuration file example should be port instead of IP

true1064:
fix(doc): the value of "profPort" in client's configuration file example should be port instead of IP


Victor1319 (approved):
LGTM

leonrayang (approved):
LGTM

xiaochunhe (commented):
LGTM

PR #2529, opened 2023-08-22
enhance(master/cli): Reduce printing too much information about replca file count and size differ:

true1064:
enhance(master/cli): Reduce printing too much information about replca file count and size differ:
(1)master: not record replica file count and size differ if datapartition is in decommission
(2)cli: only print number of datapartitions with replica file count and size differ in default; and add an argument to control if print such datapartitions's info


Victor1319 (approved):
LGTM

bboyCH4 (approved):

PR #2543, opened 2023-08-24
bugfix(master):fix the issue of master being suspended when reading dp metadata

bboyCH4:
1. add dp into decommission list by goroutine when master load meta data
2. restor e decommission progress for disk when master restart


leonrayang (approved):
LGTM

true1064 (approved):
LGTM

PR #2562, opened 2023-08-29
bugfix(master): fix deadlock when resotre running dp meta

bboyCH4:
fix deadlock when resotre running dp meta


leonrayang (approved):
LGTM

PR #2567, opened 2023-08-30
enhance(metanode).add interface of IsRestoring to raft store for bett…

leonrayang:
…er performace while do IsFollowerRead jugement

To ensure that the partition is not in a snapshot state during metanode's followRead, we queried the status interface of the raft. However, the Status() interface of raft is implemented using a channel for lock-free operation, which leads to poor performance. To address this, we are adding an interface here that allows us to retrieve only the required status instead of the entire raft state.

fixes #

**Release note**:

```release-note
```

### DIFF
--- a/depends/tiglabs/raft/server.go
+++ b/depends/tiglabs/raft/server.go
@@ -218,6 +218,15 @@ func (rs *RaftServer) ChangeMember(id uint64, changeType proto.ConfChangeType, p
 	return
 }
 
+func (rs *RaftServer) IsRestoring(id uint64) bool {
+	rs.mu.RLock()
+	defer rs.mu.RUnlock()
+	if raft, ok := rs.rafts[id]; ok {
+		return raft.restoringSnapshot.Get() || raft.applied() == 0
+	}
+	return true
+}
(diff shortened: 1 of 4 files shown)


Victor1319 (dismissed):
LGTM

depends/tiglabs/raft/server.go
```diff
@@ -218,6 +218,15 @@ func (rs *RaftServer) ChangeMember(id uint64, changeType proto.ConfChangeType, p
 	return
 }
 
+func (rs *RaftServer) IsRestoring(id uint64) bool {
+	rs.mu.RLock()
+	defer rs.mu.RUnlock()
```
xiangcai1215, line 223: rs.mu.RLock()
	raft, ok := rs.rafts[id]
	rs.mu.RUnlock()
	if ok {
		return  raft.restoringSnapshot.Get() && raft.applied() == 0
	}
这样锁友好点

Victor1319 (approved):
LGTM

xiangcai1215 (approved):
LGTM

PR #2569, opened 2023-08-30
[Fix]: `CarryWeightNodesetSelector` will be equal with `RoundRobinNodesetSelector` in specified case

NaturalSelect:

* In specified case (only 2 nodesets in a zone), current implementation of `CarryWeightNodesetSelector` will be equal with `RoundRobinNodesetSelector`.
* The pull request fix the bug and add more tests to nodeset selector.

**Release note**:

```release-note
```


leonrayang (approved):
LGTM

bboyCH4 (approved):

PR #2579, opened 2023-09-04
fix(Raft): fix the error when Raft has no leader which not handler he…

xiangcai1215:
…artbeat request

1.The RAFT node was unable to receive the null entry message that leder change will send when it restarts.As a result, this node will become PreCandidate. The Raft that has no leader will not handle heartbeat messages, so it will never change to a follower until an append request is received.

fixes #
1. Let PreCandidate handle heartbeat requests so it can quickly switch to follower if the cluster has a normal leader.

**Release note**:

```release-note
```


Victor1319 (approved):
LGTM

leonrayang (approved):
LGTM

PR #2585, opened 2023-09-04
snapshot. update process of version sequence check in strategy and update log ouput 

leonrayang:
enhance(master): snapshot. update process of version sequence check in strategy since ver update from second to microscond
enhance(Raft): update error output to warn since the mp inconstent may be caused by raft log behind

fixes #

**Release note**:

```release-note
```


Victor1319 (approved):
LGTM

baijiaruo (approved):
LGTM

PR #2594, opened 2023-09-06
fix some bugs related to snapshot

leonrayang:
enhance(metanode): check version no need to reset all mask euqual towq MultiVersionFlag
fix(master):snapshot.make WaitGroup usage clear in aync job in 2phase version managment in case of wrong number counter
fix(metanode):snapshot.Extent append by diff verseq need consider as split key and add to refMap,or else the ek drop caused by modification and be considered as normal discard , the whole extent would be dropped

fixes #

**Release note**:

```release-note
```


bboyCH4 (approved):

Victor1319 (approved):
LGTM

bboyCH4 (approved):

PR #2598, opened 2023-09-07
 snapshot. mp update version list according master and apply by raft submit

leonrayang:

enhance(metanode): snapshot. mp update version list according master and apply by raft submit
    
    1)apply to avoid conflict with version commit process
    2)consistent with master verlist then do verlist append and delete
    3)normal version commit from master also need consider the seq in the mp list


Victor1319 (dismissed):
LGTM

true1064 (approved):
LGTM

PR #2604, opened 2023-09-11
snapshot.enhance version creation and syncronization process

leonrayang:

  enhance(master):load multi version error need exit
  fix(master):version time cann't be equal
  enhance(master):snapshot.update creatiton and deletion condition in strategy
  enhance(metanode):snapshot. version message from master need raft submit and quened


bboyCH4 (dismissed):

baijiaruo (approved):
LGTM

PR #2607, opened 2023-09-12
enhance(metanode): snapshot.toplayer deletetion wrong report and do d…

leonrayang:
…elete according to version list

version list do deletion may be faster than lcnode do really deletion, once the top layer deletion happend(always tiggered by user operation), and not find the next layer,then consider the layer already the last layer, and the extents can be dropped.

fixes #

**Release note**:

```release-note
```


baijiaruo (approved):
LGTM

Victor1319 (approved):
LGTM

PR #2610, opened 2023-09-12
fix(metanode):snapshot.extent key in append need consider the probabi…

leonrayang:
…lity of split

current senario
1) seq update: updated as the volume information is fetched, and later it is updated based on changes in the verlist of data and meta. However, it does not have real-time capability. 2) The dirty flag of ExtentHandler is written back to the datanode, which means it is prepared for flushing to the metanode. 3) For appending ek, find the previous appendable ek. If ExtentHandler is not closed, the ek will include the previous ek (Note that it is difficult to control here as the previous ek might have already been asynchronously written to the meta or it may be in transit. After data is returned and before meta is returned, dirty is true). To append as a whole, where meta and data have changed but the client may not perceive it in a timely manner, the range of the appended ek is larger, including the previous ek (an extent). When meta is processed, it is used directly.

Improvements:
Approach 1: Without modifying the client, append the ek to the meta. The meta will differentiate the range based on the existing ke conditions, for example, if the new ek range overlaps with the old one, it will proactively split into two sets and assign a new version number to the newly added ek. Approach 2: If the meta detects that the client's requested seq is outdated and its own seq has advanced beyond the client's request, it will return to the client. The client can then modify its ek to be appended based on the current version information returned by the meta and the current state of eks in the meta.

conclusion:
Option 1 is relatively simple and covers more scenes.

fixes #

**Release note**:

```release-note
```

### DIFF
--- a/client/fs/super.go
+++ b/client/fs/super.go
@@ -117,7 +117,6 @@ func NewSuper(opt *proto.MountOptions) (s *Super, err error) {
 		ValidateOwner:   opt.Authenticate || opt.AccessKey == "",
 		EnableSummary:   opt.EnableSummary && opt.EnableXattr,
 		MetaSendTimeout: opt.MetaSendTimeout,
-		VerReadSeq:      opt.VerReadSeq,
 	}
 	s.mw, err = meta.NewMetaWrapper(metaConfig)
 	if err != nil {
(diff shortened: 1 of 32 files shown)


Victor1319 (approved):
LGTM

baijiaruo (approved):
LGTM

PR #2631, opened 2023-09-21
fix(master):snapshot.dead lock in process of check strategy and creat…

leonrayang:
…e version

resolve dead lock in process of check  snapshot strategy


true1064 (approved):

Victor1319 (approved):
LGTM

PR #2640, opened 2023-09-22
fix(master): fix the bug of not validating the cluster name when call…

### LINKED ISSUE #2639: [Bug]: When calling loadClusterValue, need to verify whether the current cluster name matches the cluster name in the rocksdb record.
### Contact Details

_No response_

### Priority

low (Default)

### Environment

```markdown
- CubeFS version:3.3.0
- Deployment mode(docker or standalone or cluster):
- Dependent components:
- OS kernel version(Ubuntu or CentOS):
- CPU/Memory:
- Others:
```

### Current Behavior

If the cluster name is changed and the records in rocksdb are not cleaned up, then the call to loadClusterValue may save the uncleaned cluster configuration to the current cluster configuration.

### Expected Behavior

When loadClusterValue is called, it will check if the cluster name recorded in rocksdb matches the current cluster name, and if it does not, it will be recorded in the log.

### Steps To Reproduce

_No response_

### CubeFS Log

_No response_

### Anything else? (Additional Context)

_No response_

shuqiang-zheng:
…ing loadClusterValue.

If the cluster name has been changed and the records in rocksdb have not been cleaned up, the cluster configuration that has not been cleaned up may be saved to the current cluster configuration when loadClusterValue is called, so  it is necessary to determine whether the cluster name matches or not when loadClusterValue is called.

fixes #2639 

**Release note**:

```release-note
```

### DIFF
--- a/master/metadata_fsm_op.go
+++ b/master/metadata_fsm_op.go
@@ -1035,6 +1035,11 @@ func (c *Cluster) loadClusterValue() (err error) {
 			return err
 		}
 
+		if cv.Name != c.Name {
+			log.LogErrorf("action[loadClusterValue] loaded cluster value: %+v", cv)
+			continue
+		}
+
 		log.LogDebugf("action[loadClusterValue] loaded cluster value: %+v", cv)
 		c.CreateTime = cv.CreateTime
 


Victor1319 (approved):
LGTM

leonrayang (approved):
LGTM

PR #2643, opened 2023-09-25
fix(client): snapshot.try append ek process forget the discards process and check failed in metanode

leonrayang:

1) try append forget the discards process and check failed in metanode
    write request cover the old ek and do append write for performace and discard old ek in the flush stage,
    but the try append(client viewpoint don't know whether the extent right neighbor exist or not,need to try write,
    but the async process of do append write need async dealing, which not easy to manage,
    so use direct write append by raft submit in datanode for simple)
    process use sync way but forget the discard process lead to inconsistent in metanode check
2) defer unlock is stale code and clean or else lead to deadlock


bboyCH4 (approved):

baijiaruo (approved):
LGTM

PR #2645, opened 2023-09-25
fix(master):last deadlock not complete

leonrayang:

the last commit clean the defer lock ,but forget release at the end if condition not satisfied and lead to deadlock


baijiaruo (approved):
LGTM

bboyCH4 (approved):

Issue #2019, opened 2023-05-25
[Bug]: create volume init data error but can get volume


xiangcai1215:
**Describe the bug**
An init data partition error results in insufficient data partitions in a volume's partition map. However, the volume already exists in the system, so it can be found in the master.

**Environment**
Describe your server system version, kernel version

**To Reproduce**

**Expected behavior**

**Additional context**

true1064:
There seems to be a problem

true1064:
@xiangcai1215  plz check this out, seems you have fixed it?

true1064:
confirmed that it is a bug. plan to fix it in develop-3.4.0

PR #2653, opened 2023-09-27
fix(master): if init data partitions failed when create volume, mark volume status as delete to trigger background deletion.

true1064:
currently, if initDataPartitions() failed in the creating vol procedure, the vol info is not cleaned, so it can be found in the master but can't be used.
should clean the vol info if initDataPartitions() failed when creating.

Which issue this PR fixes:

fixes https://github.com/cubefs/cubefs/issues/2019


Victor1319 (approved):
LGTM

PR #2668, opened 2023-10-08
refactor(client): if request datanode to create extent failed, not remove the datapartition form dpSelector if it is the last one.

true1064:
if request datanode to create extent failed, not remove the datapartition form dpSelector if it is the last one，
to avoid the dpSelector has no datapartition .

if any dn host of the dp can not be connected, it will be add to the exclude map and not retry:
```
func (dp *DataPartition) CheckAllHostsIsAvail(exclude map[string]struct{}) {
	var (
		conn net.Conn
		err  error
	)
	for i := 0; i < len(dp.Hosts); i++ {
		host := dp.Hosts[i]
		if conn, err = util.DailTimeOut(host, proto.ReadDeadlineTime*time.Second); err != nil {
			log.LogWarnf("CheckAllHostsIsAvail: dial host (%v) err(%v)", host, err)
			if strings.Contains(err.Error(), syscall.ECONNREFUSED.Error()) {
				exclude[host] = struct{}{}
			}
			continue
		}
		conn.Close()
	}
}
```

### DIFF
--- a/sdk/data/wrapper/data_partition_selector.go
+++ b/sdk/data/wrapper/data_partition_selector.go
@@ -38,6 +38,9 @@ type DataPartitionSelector interface {
 
 	// RemoveDP removes specified data partition.
 	RemoveDP(partitionID uint64)
+
+	// Count return number of data partitions held by selector.
+	Count() int
 }
 
 var (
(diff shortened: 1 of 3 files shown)


leonrayang (approved):
LGTM

Victor1319 (approved):
LGTM

PR #2669, opened 2023-10-08
snapshot.clean tempory snapshot by inner periodic check

leonrayang:

 enhance(metanode):snapshot.clean tempory snapshot by inner periodic check
    
Adopting a one-phase commit means that the master processor (MP) considers the prepared version as valid, and during the operation, it treats it as a version. This means that the MP may have partial temporary version information. However, the final valid version is on the master, which means that the dependency for reading historical versions on the edge side is the version list of the master, not the version list of the MP. When attempting versions in the form of 0, 100, 150, 200, with 150 being a temporary version, the version 100 is read, and it needs to include 150; otherwise, some data cannot be read due to the temporary version 150. Therefore, on the client side, it is necessary to read using the formal version's seq-1 method, which means reading version 199.
    
 Question: The MP is distributed with versions, some of which have temporary versions and some do not. How should the cleanup of temporary versions be handled? If the dentry is in a partition with temporary versions, but the inode is not, when reading the dentry and obtaining the temporary version, is it possible for the inode data to be misaligned?
    
1) For example, file creation in version 0, creating temporary version 50, with the dentry in the MP being effective, but the inode partition not being effective. At this point, if the dentry is deleted, the retained temporary version 50's dentry information remains, but is it possible that the inode has already been deleted, resulting in an exception when reading with temporary version 50?
    
2) For rename operations, referring to the situation described in 1), as long as it is not read, it will not affect the final state, and version cleanup should also not affect the previous state.
    
    Analysis and answer: 1) User operations will not delete the inode because the dentry must be deleted first. 2) When cleaning up temporary versions in snapshots, the inode will not be deleted because inode version cleanup only considers intermediate versions. 3) When reading, the edge side reads the next formal version minus one, so it sees a consistent version. Therefore, there is no need for targeted processing in this case.

### DIFF
--- a/metanode/dentry.go
+++ b/metanode/dentry.go
@@ -106,6 +106,7 @@ func (d *Dentry) isDeleted() bool {
 func (d *Dentry) setDeleted() {
 	if d.multiSnap == nil {
 		log.LogErrorf("action[setDeleted] d %v be set deleted not found multiSnap", d)
+		return
 	}
 	log.LogDebugf("action[setDeleted] d %v be set deleted", d)
 	d.multiSnap.VerSeq |= uint64(1) << 63
--- a/metanode/extend.go
+++ b/metanode/extend.go
@@ -17,6 +17,7 @@ package metanode
 import (
 	"bytes"
 	"encoding/binary"
+	"fmt"
 	"sync"
 
 	"github.com/cubefs/cubefs/util/btree"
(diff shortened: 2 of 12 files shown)


Victor1319 (approved):
LGTM

true1064 (approved):

PR #2675, opened 2023-10-10
fix(docs): fix the problem that version.json format is incorrect

baijiaruo:
 fix the problem that version.json format is incorrect


true1064 (approved):
LGTM

Victor1319 (approved):
LGTM

PR #2684, opened 2023-10-17
feat(master): support max mp split when max mp is no leader

tangdeyi:
1. support max mp split when max mp is no leader.
2. modify lowerLimitRWMetaPartition to 3，which equal defaultReplicaNum when volume was created.


yhjiango (approved):
LGTM

honeyvinnie (approved):
LGTM

PR #2685, opened 2023-10-18
fix(master,datanode):Initialize Apply file and add a check condition for dp rollback during dp creation

bboyCH4:
1.Apply will be initialized when new datapartition is created.
2.Only sync up metadate of datapartitons when the creation of the datapartitions fails in the decommission process.


Victor1319 (approved):
LGTM

baijiaruo (approved):
LGTM

PR #2691, opened 2023-10-20
fix(objectnode): fix read body entirely into memory and without an up…

tangdeyi:
fix read body entirely into memory


honeyvinnie (approved):
LGTM

yhjiango (approved):
LGTM

PR #2711, opened 2023-10-25
fix(metanode): snapshot.verlist last seq should not consider as tempo…

leonrayang:
…ry verSeq and do deletion

verlist last seq should not consider as tempory verSeq and do deletion, Otherwise, if the seq of the MP rolls back, it will result in abnormal behavior after metanode restart and do replay.

enhance xattr snapshot management

add log for master critical management


Victor1319 (approved):
LGTM

baijiaruo (approved):
LGTM

PR #2712, opened 2023-10-26
fix(metanode):truncate refMap not be set trigger panic while restart

leonrayang:

truncate refMap not be set trigger panic while restart，truncate punch hole on extent and it's extent key should add to ref map, otherwise may trigger panic.


Victor1319 (approved):
LGTM

true1064 (approved):
LGTM

PR #2721, opened 2023-10-30
fix(client):snapshot. direct write need update handler's fileoffset a…

leonrayang:
…nd size

The "try direct write" was not performed using streaming write, but the subsequent extent handler's key was not updated. Additionally, this key may have changed its sequence due to the direct write attempt. As a result, when attempting the next append write, it does not meet the conditions for direct append write because the offset of the handler's key is not updated. Furthermore, when creating a new handler, the original handler needs to be closed. However, the handler retains information from the key before the "try direct write," and since the sequence has already changed, the extent cache is updated and replaced with the new sequence. This leads to subsequent writes mistakenly assuming that direct modification write is possible.


bboyCH4 (approved):

bboyCH4 (approved):

Victor1319 (approved):
LGTM

Issue #2481, opened 2023-08-11
[Feature]: Support delete volume only when dentry count is zero.


Victor1319:
**Is your feature request related to a problem? Please describe.**
 Currently, volumes support direct deletion, even if there is still data inside the volume. There is a high risk of accidental deletion.

**Describe the solution you'd like**
The ability to delete a volume is supported only when the number of dentries inside the volume is zero.

true1064:
good idea

NaturalSelect:
This problem still exists?

Victor1319:
yes, still exist @NaturalSelect

NaturalSelect:
> yes, still exist @NaturalSelect

Hello @Victor1319.

My understanding is that the volume's dentry count is checked by the Master when mark delete volume.

Please point out any misunderstandings on my part.

Victor1319:
yes, before delete volume, check dentry count whether zero or not on master.

chenzufei:
I would like to participate in resolving this issue.

jiangbbaoyu:
This is my solution,  fetch latest dentry count of the deleting volume from metaNode , if dentry count greater than zero , abort the delete operation   https://github.com/cubefs/cubefs/pull/2724.  My first try on cubeFS, any problem please contact me.  @Victor1319

jaxjiangg:
resolved ?

PR #2724, opened 2023-10-31
fix [Feature]: Support delete volume only when dentry count is zero.

jiangbbaoyu:
Check latest dentry count of the deleting volume, if dentry count greater than zero , abort the delete operation.

fixes # https://github.com/cubefs/cubefs/issues/2481

**Release note**:

```release-note
```

### DIFF
--- a/master/cluster.go
+++ b/master/cluster.go
@@ -1289,23 +1289,24 @@ func (c *Cluster) markDeleteVol(name, authKey string, force bool) (err error) {
 		return proto.ErrVolNotExists
 	}
 
-	vol.mpsLock.RLock()
-	volDentryCount := uint64(0)
-	for _, mp := range vol.MetaPartitions {
-		// to avoid latency, fetch latest mp dentry count from metanode
-		c.doLoadMetaPartition(mp)
-		mpDentryCount := uint64(0)
-		for _, response := range mp.LoadResponse {
-			if response.DentryCount > mpDentryCount {
-				mpDentryCount = response.DentryCount
+	if !force {
+		volDentryCount := uint64(0)
+		mpsCopy := vol.cloneMetaPartitionMap()
+		for _, mp := range mpsCopy {
+			// to avoid latency, fetch latest mp dentry count from metanode
+			c.doLoadMetaPartition(mp)
+			mpDentryCount := uint64(0)
+			for _, response := range mp.LoadResponse {
+				if response.DentryCount > mpDentryCount {
+					mpDentryCount = response.DentryCount
+				}
 			}
+			volDentryCount += mpDentryCount
 		}
-		volDentryCount += mpDentryCount
-	}
-	vol.mpsLock.RUnlock()
 
-	if volDentryCount > 0 {
-		return fmt.Errorf("vol %s is not empty ! dentry count: %d ", vol.Name, volDentryCount)
+		if volDentryCount > 0 {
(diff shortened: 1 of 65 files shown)


NaturalSelect:
Consider adding unit tests for it?

NaturalSelect:
> > Consider adding unit tests for it?
> 
> unit test is required or optional ? for this unit test case , i need to fake a master/meta/data processes, but i can't find any unit test infrastructure code i can use

I can find other unit tests in `api_service_test.go`.

jiangbbaoyu:
> > > Consider adding unit tests for it?
> > 
> > 
> > unit test is required or optional ? for this unit test case , i need to fake a master/meta/data processes, but i can't find any unit test infrastructure code i can use
> 
> I can find other unit tests in `api_service_test.go`.

thanks, i saw them

master/cluster.go
```diff
@@ -1289,6 +1289,25 @@ func (c *Cluster) markDeleteVol(name, authKey string, force bool) (err error) {
 		return proto.ErrVolNotExists
 	}
 
+	vol.mpsLock.RLock()
+	volDentryCount := uint64(0)
+	for _, mp := range vol.MetaPartitions {
```
Victor1319, line 1294: your lock cost maybe too bigger, which may casue other request.
here, you can use the backup of vol.metapartitions, eg: 
`vol.mpsLock.RLock()
newMap := vol.cloneMetaPartitions.
vol.mpsLock.RUnLock()
`

master/cluster.go
```diff
@@ -1289,6 +1289,25 @@ func (c *Cluster) markDeleteVol(name, authKey string, force bool) (err error) {
 		return proto.ErrVolNotExists
 	}
 
+	vol.mpsLock.RLock()
```
Victor1319, line 1292: it's better to add a switch config to control whether to check `dentryCnt == 0`, eg, add a config key `forceDelVol`, when it's true, not check dentry count,  `false` will check. And set default value is true to keep consistent with the original deletion plan.
And it's also will be better to add a config key like `delDentryCnt`, when dentyCnt < `delDentryCnt`, vol can be deleted.

master/cluster.go
jiangbbaoyu: good suggestion ! i will fix soon

master/cluster.go
jiangbbaoyu: if default value of forceDelVol  is true , risk of accidental deletion still exists.  I think false is the resonable default value.   And i will fix the  ci-test-unit and ci-test-s3 error

Victor1319: In fact, it will be better to add a switch in the configuration file of master, when the switch is turned on, the entire volume can be deleted directly, and when it is not opened, only the volume with the number of dentries less than the specified value can be deleted, which is more secure and avoids misoperation.

jiangbbaoyu: shoud i implement this requirements in this issue ?
1) add _volFroceDeletion_ in master configuration file (default false)
2) add _volDeletionDentryThreshold_ in master configuration file (default 10)

if so  i will undo the cfs-cli  and master AdminDeleteVol interface changes

Victor1319: yes, it's better to implement the both requirements, but `volFroceDelete` can be set default true, so it will keep consistent with the versions before.

jiangbbaoyu: i reopened a pr to resolve this

PR #2732, opened 2023-10-31
refactor(data): optimize the determination of datanode disk's unavailable status:

true1064:
refactor(data): optimize the determination of datanode disk's unavailable status:
(1) add 2 conf item: diskUnavailableErrorCount and diskUnavailablePartitionErrorCount. 
(2) when encounter disk error, set disk as unavailable if disk error count reaches diskUnavailableErrorCount or partition count with disk error reaches diskUnavailablePartitionErrorCount. 
(3) if a partition encounter 1 disk error, set the partition's status as unavailable.


bboyCH4 (approved):

Victor1319 (approved):
LGTM

PR #2735, opened 2023-11-01
bugfix(datanode,cli,datanode): fix some logic for dp decommission

bboyCH4:
1.remove logic of fetching replica from master after adding raft member
2.cli show error msg when decommission failed 3.keep disk original decommission status when load from rocksdb
3. if no disk to decommission,datanode reset to initial 2. if dp reach max retry times, delete the added replica and add the source replica.


Victor1319 (dismissed):
LGTM

true1064 (approved):

Victor1319 (approved):
LGTM

true1064 (approved):

PR #2736, opened 2023-11-01
fix(util): check if err occurred when get CPU percent, to avoid panic if got nil result.

true1064:
fix(util): check if err occurred when get CPU percent, to avoid panic if got nil result.

### DIFF
--- a/util/loadutil/cpu.go
+++ b/util/loadutil/cpu.go
@@ -15,12 +15,30 @@
 package loadutil
 
 import (
+	"fmt"
+	"github.com/cubefs/cubefs/util/log"
 	"time"
 
 	"github.com/shirou/gopsutil/cpu"
 )
 
-func GetCpuUtilPercent(sampleDuration time.Duration) float64 {
-	utils, _ := cpu.Percent(sampleDuration, false)
-	return utils[0]
+func GetCpuUtilPercent(sampleDuration time.Duration) (used float64, err error) {
+	utils, err := cpu.Percent(sampleDuration, false)
+	if err != nil {
+		log.LogErrorf("[GetCpuUtilPercent] err: %v", err.Error())
+		return
+	}
+	if utils == nil {
+		err = fmt.Errorf("got nil result")
+		log.LogErrorf("[GetCpuUtilPercent] err: %v", err.Error())
+		return
+	}
+	if len(utils) == 0 {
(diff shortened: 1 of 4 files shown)


util/loadutil/cpu.go
```diff
@@ -15,12 +15,30 @@
...
 
 	"github.com/shirou/gopsutil/cpu"
 )
 
-func GetCpuUtilPercent(sampleDuration time.Duration) float64 {
-	utils, _ := cpu.Percent(sampleDuration, false)
-	return utils[0]
+func GetCpuUtilPercent(sampleDuration time.Duration) (used float64, err error) {
+	utils, err := cpu.Percent(sampleDuration, false)
+	if err == nil {
```
NaturalSelect, line 27: `if err != nil`

true1064: fixed

util/loadutil/cpu.go
```diff
@@ -15,12 +15,30 @@
...
-func GetCpuUtilPercent(sampleDuration time.Duration) float64 {
-	utils, _ := cpu.Percent(sampleDuration, false)
-	return utils[0]
+func GetCpuUtilPercent(sampleDuration time.Duration) (used float64, err error) {
+	utils, err := cpu.Percent(sampleDuration, false)
+	if err != nil {
+		log.LogErrorf("[GetCpuUtilPercent] err: %v", err.Error())
+		return
+	}
+	if utils == nil {
```
NaturalSelect, line 31: This part is unnecessary.

If `utils` is nil, `len(utils)` will return `0`.

NaturalSelect: See also: [doc](https://pkg.go.dev/builtin#len).

true1064: it's for this case: if the utils ni not nil but its len is 0

cpu.Percent( ) may be modified in the future， this check may prevent possible exceptions

NaturalSelect: > it's for this case: if the utils ni not nil but its len is 0
> 
> cpu.Percent( ) may be modified in the future， this check may prevent possible exceptions

We may just need `if len(utils) == 0`?

baijiaruo (approved):
LGTM

Victor1319 (approved):
LGTM

PR #2744, opened 2023-11-04
fix [Feature]: Support delete volume only when dentry count is zero.

jiangbbaoyu:
if volForceDeletion param set to true, volume can be deleted directly. otherwise, volumes can be deleted only if it's latest dentry count is less than or equal to the volDeletionDentryThreshold

fixes # https://github.com/cubefs/cubefs/issues/2481

**Release note**:

```release-note
```


Victor1319 (approved):
LGTM

true1064 (approved):
LGTM

PR #2756, opened 2023-11-09
bugfix(datanode): fix datanode raft snapshot exceed the limit

bboyCH4:
If the DP offline operation triggers Raft to synchronize WAL logs through snashot, it will cause an infinite loop when generating the snashot iterator


Victor1319 (dismissed):
LGMT

PR #2757, opened 2023-11-09
feat(data): add an API to manually set a disk of dataNode as error. t…

true1064:
feat(data): add an API to manually set a disk of dataNode as error.
 this operation will set disk status as unavailable and stop raft of all data partitions on the disk


Victor1319 (approved):
LGTM

baijiaruo (approved):
LGTM

Issue #1909, opened 2023-04-17
[Feature]: Process (all) log storage rolls according to directory space


pengtianyue025:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
- Support for configuring the number of logs to keep
- Support for configuring log rotation size
- The log automatically scrolls when there is not enough space

**Additional context**
The directory space is not taken into account when the log is currently written, and automatic scrolling cannot be implemented according to the directory to which it belongs. If the space is full, the main interface will be hung and the main process will be affected.

setcy:
Hello， I want to try this issue

PR #2762, opened 2023-11-11
Support config rotate size and head room 

setcy:

- unify 'rotate' and 'rolling' into 'rotate'
- **support config rotate size and head room**

fixes: #1909

It seems 'log automatically scrolls when there is not enough space' already has been implemented.


sejust (approved):

Victor1319 (approved):
LGTM

PR #2764, opened 2023-11-13
fix(objectnode): s3api perm is compatible with fuse perm

tangdeyi:
s3api perm is compatible with fuse perm


honeyvinnie (approved):
LGTM

yhjiango (approved):
LGTM

PR #2806, opened 2023-11-27
Fix snapshot inconsistent bugs about modification based on snapshot

leonrayang:
1）extent cache wrongly update
2）during lastKey reuse handler may not have a effective key
3)  datanode mod append write allocated wrongly


bboyCH4 (approved):

Victor1319 (approved):
LGTM

PR #2848, opened 2023-12-08
bugfix(raft): fix a bug abount raft decode operation

bboyCH4:
When the length of the data is 0, return directly without further decoding


Victor1319 (approved):
LGTM

mawei029 (approved):

PR #2863, opened 2023-12-11
bugfix(master): check decommission disk/data node when meta is ready

bboyCH4:
1.modify node selector log level 
2. only check decommission disk/data node when meta is ready


Victor1319 (approved):
LGTM

true1064 (approved):

PR #2878, opened 2023-12-13
fix some bugs about wrong deletion duiring snapshot deletion and modification

leonrayang:

fix(metanode):append key in sequnce jugement didn't include extent id
fix(metanode):snapshot. dentry deletion on top layer with no snapshotlist then update toplayer to deletion and store version to list althrough no version in mp verlist
    
    version of inode exist identify that it's snapshot version not do deletion  on this node, no matter mplist exist or not, because mpVerlist update while recive master's command of
    version deletion,but real deletion may do deletion aysncronus and may not reach the inode, but user can do deletion on top layer at the same time.


Victor1319 (approved):
LGTM

true1064 (approved):

bboyCH4 (approved):

PR #2882, opened 2023-12-14
bugfix(metanode):check sortedEks is nil when excute fsmAppendExtentsW…

bboyCH4:
…ithCheck
check sortedEks is nil when excute fsmAppendExtentsWithCheck


honeyvinnie (approved):

true1064 (approved):
LGTM

PR #2884, opened 2023-12-15
[fix]:fix mismatching between information of master and datanode bug …

zhang111feng:
…when adding replica

Fix mismatching between information of master and datanode bug when adding replica. The order in which asynchronous goroutines are executed can cause this bug.If we execute addDataPartitionRaftMember() -> buildAddDataPartitionRaftMemberTaskAndSyncSendTask() -> ChangeRaftMember() -> ApplyMemberChange() -> updateReplicas() first, then execute addDataPartitionRaftMember(){append hosts;append peers},which will result in datanode fetching old information of dp.replica from master and replacing the latest information included newly added replica with the old information excluded newly added replica

null

null

**Release note**:

```release-note
```

### DIFF
--- a/master/cluster.go
+++ b/master/cluster.go
@@ -2530,6 +2530,26 @@ func (c *Cluster) addDataPartitionRaftMember(dp *DataPartition, addPeer proto.Pe
 		return nil
 	}
 
+	dp.Lock()
+
+	oldHosts := make([]string, len(dp.Hosts))
+	copy(oldHosts, dp.Hosts)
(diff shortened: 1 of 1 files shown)


Victor1319:
it's may be better and easy to sleep some time befor invoke updateReplicas() func.

zhang111feng:
> it's may be better and easy to sleep some time befor invoke updateReplicas() func.

That's fine,but how can I determine the precise time of how long it should sleep before update executes.

PR #2902, opened 2023-12-19
bugfix(client):only clear dp when volume storage class is blobstore

bboyCH4:

bugfix(client):only clear dp when volume storage class is blobstore

### DIFF
--- a/cli/cmd/fmt.go
+++ b/cli/cmd/fmt.go
@@ -155,7 +155,7 @@ func formatSimpleVolView(svv *proto.SimpleVolView) string {
 	for _, asc := range svv.AllowedStorageClass {
 		allowedStorageClassStr = append(allowedStorageClassStr, proto.StorageClassString(asc))
 	}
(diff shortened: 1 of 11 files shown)


proto/admin_proto.go
```diff
@@ -1230,13 +1230,8 @@ func IsStorageClassBlobStore(storageClass uint32) bool {
...
 }
 
-func VolSupportsBlobStore(allowedStorageClass []uint32) bool {
-	for _, storageClass := range allowedStorageClass {
-		if storageClass == StorageClass_BlobStore {
-			return true
-		}
-	}
-	return false
+func VolSupportsBlobStore(storageClass uint32) bool {
```
true1064: 这个函数不用改。用原来的proto.IsStorageClassBlobStore(volumeInfo.VolStorageClass)

true1064 (approved):

honeyvinnie (approved):

PR #2910, opened 2023-12-20
client append ek which seq may be updated by metanode,add ref for consistent

leonrayang:

    in append ek process,if do neigbor ek merge together of the same extent,the client may have different view of eks,
    then the client opertion will not consistent with metanode,and may trigger conflicts.so just add ref for this scenrio,
    because the verseq update not always happen


bboyCH4 (approved):

Victor1319 (approved):
LGTM

PR #2918, opened 2023-12-21
fix(master)：fix the calculation of vol usedSize of API /client/volStat, not set usedSize as totalSize when usedSize > totalSize

true1064:
…t, not set usedSize as totalSize when usedSize > totalSize

fix(master)：fix the calculation of vol usedSize of API /client/volStat, not set usedSize as totalSize when usedSize > totalSize.
This will show the right used size for volumes whose usedSize is larger than totalSize


bboyCH4 (approved):

leonrayang (approved):
LGTM

PR #2919, opened 2023-12-21
fix(client): set usedSize as totalSize if usedSize > totalSize when queried from master, to avoid displaying anomalies such as cmd df

true1064:
fix(client): set usedSize as totalSize if usedSize > totalSize when queried from master, to avoid displaying anomalies such as cmd df


leonrayang (approved):
LGTM

leonrayang (approved):
LGTM

shuqiang-zheng (approved):

PR #2924, opened 2023-12-22
bugfix(client,master):fix the calculation of vol usedSize

bboyCH4:
fix the calculation of vol usedSize


leonrayang (approved):
LGTM

mawei029 (approved):

PR #2951, opened 2023-12-27
fix snapshot bugs: 1)extent meta reference management optimization 2)snapshot deletion

leonrayang:

fix(metanode):Snapshot.refactor inode deletion and consider system verlist dependency
fix(metanode): Regarding reference counting, performing append writes in the middle of the EKS and unifying the condition for appending as the last key (lastKey) of the EKS


true1064 (approved):

Victor1319 (approved):
LGTM

shuqiang-zheng (approved):

PR #2968, opened 2024-01-04
feat(master): [hybrid cloud] handle compatibility of master when older version upgrade to hybrid cloud

true1064:
feat(master): [hybrid cloud] handle compatibility of master when older version upgrade to hybrid cloud
(1) auto set mediaType of datanode, zone, datapartiton by config legacy "legacyDataMediaType" and persist; 
(2) auto set storageClass of volume by config legacy "legacyDataMediaType" and persist; 
(3) if need to rollback master to older version, need to do nothing, just replace to old master and restart. because hybrid cloud master only add new fields to meta by not changed old fields.


honeyvinnie (approved):
LGTM

PR #2980, opened 2024-01-09
feat(data): if IO errors make disk status change to unavailable, not set all datapartitions of the disk as unavailable except the datapartition related to the if IO errors.

true1064:
feat(data): if IO errors make disk status change to unavailable, not set all datapartitions of the disk as unavailable except the datapartition related to the if IO errors.


Victor1319 (approved):
LGTM

leonrayang (approved):
LGTM

Victor1319 (approved):
LGTM

PR #2984, opened 2024-01-10
fix(data): fix statusUpdate() to rightly set datapartition status

true1064:
fix(data): fix statusUpdate() to rightly set datapartition status


Victor1319 (approved):
LGTM

Issue #2738, opened 2023-11-02
[Bug]:  datanode raft send snapshot will block once snapshot operation occurs.


Victor1319:
### Is there an existing issue for this?

- [X] I have searched all the existing issues

### Priority

high

### Environment

```markdown
- CubeFS version: master
- Deployment mode(docker or standalone or cluster): all
- OS kernel version(Ubuntu or CentOS): centos
- CPU/Memory:
- Others:
```

### Current Behavior

when decommission datanode, it maybe block datapartition raft op.

### Steps To Reproduce

```markdown
1. create many datapartition more the 10 for every datanode.
2. make a lot of rand write and overwrite to trigger raft op for datapartition.
3. decommission one datanode.
```

### CubeFS Log

```2023-11-02 09:35:00,509 transport_replicate.go:111: [ERROR] [Transport] 140729 send snapshot to 868 failed error is: snapshot concurrency exceed the limit 10, now 11.^M
2023-11-02 09:35:00,509 transport_replicate.go:111: [ERROR] [Transport] 139670 send snapshot to 866 failed error is: snapshot concurrency exceed the limit 10, now 11.^M
2023-11-02 09:35:00,509 transport_replicate.go:111: [ERROR] [Transport] 139596 send snapshot to 866 failed error is: snapshot concurrency exceed the limit 10, now 12.^M
2023-11-02 09:35:00,509 transport_replicate.go:111: [ERROR] [Transport] 139779 send snapshot to 866 failed error is: snapshot concurrency exceed the limit 10, now 11.^M
2023-11-02 09:35:00,509 transport_replicate.go:111: [ERROR] [Transport] 140250 send snapshot to 864 failed error is: snapshot concurrency exceed the limit 10, now 11.^M
2023-11-02 09:35:00,509 transport_replicate.go:111: [ERROR] [Transport] 139721 send snapshot to 863 failed error is: snapshot concurrency exceed the limit 10, now 1

### Anything else? (Additional Context)

_No response_

PR #3002, opened 2024-01-16
bugfix(datanode):fix datanode raft snapshot exceed the limit

bboyCH4:
fix datanode raft snapshot exceed the limit
close: #2738


Victor1319 (approved):
LGTM

leonrayang (approved):

Issue #3004, opened 2024-01-17
[Enhancement]: support multiple datanode/metanode instances using different port on a single node 


morphes1995:
### Contact Details

morphes1995@gmail.com

### Is there an existing issue for this?

- [x] I have searched all the existing issues

### What would you like to be added?

For a specified data partition or meta partition , currently, the replica members must use the same raft heartbeat port and replica port, otherwise replicas will not be able to communicate with each other.

```go
func (mp *metaPartition) startRaft() (err error) {
  heartbeatPort, replicaPort, err = mp.getRaftPort() // all peers of this mp use same port
	for _, peer := range mp.config.Peers {
		rp := raftstore.PeerAddress{
			HeartbeatPort: heartbeatPort,
			ReplicaPort:   replicaPort,
		}
		peers = append(peers, rp)
	}
	pc := &raftstore.PartitionConfig{
		Peers:   peers,
	}
	mp.raftPartition, err = mp.config.RaftStore.CreatePartition(pc)
  return 
}
```

### Why is this needed?

For a single node that only has one net card , in order to deploy cubeFS, we must fake some secondary IPs to satisfy the replicas  port requirement. 

### Anything else?

None yet

morphes1995:
Another reason for this feature,  we can build a even more fully-fectured unit test system in which dn/mn not just start a http server

xiaochunhe:
> Another reason for this feature, based on which we can build a even more fully-fectured unit test system in which dn/mn not just start a http server

good advice :) , there was a PR (https://github.com/cubefs/cubefs/pull/3016) recently to solve this problem

PR #3016, opened 2024-01-18
feat(port): support multiple datanode/metanode instances using dif…

zhang111feng:
…ferent port on a single node

Support multiple datanode/metanode instances using different port on a single node.In the previous version, every datanode or metanode in the same raft group is required to be configured with the same raftHeartBeatPort and replicaPort because each node splices ports of peers through its own port.

fixes https://github.com/cubefs/cubefs/issues/3004

**Release note**:

```release-note
```

### DIFF
--- a/authnode/authnode_manager.go
+++ b/authnode/authnode_manager.go
@@ -63,6 +63,13 @@ func (m *Server) handlePeerChange(confChange *proto.ConfChange) (err error) {
 	case proto.ConfRemoveNode:
 		m.raftStore.DeleteNode(confChange.Peer.ID)
 		msg = fmt.Sprintf("clusterID[%v] peerID:%v,nodeAddr[%v] has been removed", m.clusterName, confChange.Peer.ID, addr)
+	case proto.ConfUpdatePeer:
+		var arr []string
+		if arr = strings.Split(addr, colonSplit); len(arr) < 2 {
+			msg = fmt.Sprintf("action[handlePeerChange] clusterID[%v] nodeAddr[%v] is invalid", m.clusterName, addr)
+			break
+		}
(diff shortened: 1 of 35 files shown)


Victor1319:
It's a really good and big job, and maybe you should consider the process of upgrading an old cluster because the default values for `raftport` and `replica port` are 0 when the older nodes were registered. This could potentially cause issues.

zhang111feng:
> It's a really good and big job, and maybe you should consider the process of upgrading an old cluster because the default values for `raftport` and `replica port` are 0 when the older nodes were registered. This could potentially cause issues.

I have made some attempts to resolve the compatibility issues when upgrading the old version.I have finished the dataNode part, but I think it still have problems.I think I need some help and suggestions for improvement.

zhang111feng:
The problem I encounter now is that after dataNodes are restarted one by one, although the META information in the locally stored dp file can be updated to the new version with port information, the META information may not be correctly changed when other member changes such as del-replica are performed.

zhang111feng:
I found that in the branch &lt;develop-hybridcloud-lifecycle&gt;, the branch also changes the structure of dataNode with adding `mediaType` when it adds the storage tiering function. How does that solve the problem in the process of upgrading an old cluster?

zhang111feng:
I think the problem mentioned doesn't exist anymore, but I'm not sure if there are other problems because I haven't fully understood the mechanism here.

true1064:
> The problem I encounter now is that after dataNodes are restarted one by one, although the META information in the locally stored dp file can be updated to the new version with port information, the META information may not be correctly changed when other member changes such as del-replica are performed.

true1064:
> I found that in the branch <develop-hybridcloud-lifecycle>, the branch also changes the structure of dataNode with adding `mediaType` when it adds the storage tiering function. How does that solve the problem in the process of upgrading an old cluster?
 in branch <develop-hybridcloud-lifecycle>,  filed  `mediaType` is only on the master, there no filed added on datanode

PR #3027, opened 2024-01-23
During the truncate process, the extent cache is forcefully refreshed, but it does not have any effect

leonrayang:
…k reentrancy and other lock concurrency

During the truncate process, the extent cache is forcefully refreshed, but it does not have any effect


sejust (approved):

Victor1319 (approved):
LGTM

PR #3041, opened 2024-01-25
bugfix(raft): When the length of the data is 0, return directly ...

bboyCH4:
without further decoding
When the length of the data is 0, return directly without further decoding


leonrayang (approved):
LGTM

true1064 (approved):

PR #3051, opened 2024-01-29
amend that server.user.clearUserStore() may be invoked before server.user is initialized, in func handleLeaderChange(), to avoid panic

true1064:
amend that server.user.clearUserStore() may be invoked before server.user is initialized, in func handleLeaderChange(), to avoid panic


Victor1319 (dismissed):
LGTM

shuqiang-zheng (approved):
LGTM

Victor1319 (approved):
LGTM

PR #3060, opened 2024-02-01
bugfix(client): catch the error when the ExtentHandler encounters a p…

bboyCH4:
…anic during the write operation
 catch the error when the ExtentHandler encounters a panic during the write operation


Victor1319 (approved):
LGTM

PR #3070, opened 2024-02-04
fix(client):  snapshot issue trigger by version update && update docs

leonrayang:
When processing the return from the datanode, if the sequence changes and the extent key is no longer consecutive,
the key inside the extent handler needs to be initialized.The current initialization only changes the file offset and nothing else,
which actually pollutes the extent cache and leads to anomalies in subsequent modifications based on this cache information.


Victor1319 (approved):
LGTM

shuqiang-zheng (approved):

PR #3072, opened 2024-02-04
fix(datanode): mark the extent as IsDeleted if it exist in deleteExtents map

yansenliang:
Mark the extent as 'IsDeleted' if it exist in deleteExtents map，no need to add it to deleteExtents.

fixes #

**Release note**:

```release-note
```


leonrayang:
Thanks @yansenliang for this datanode extent fix!

We're closing this PR because it has been open since early 2024 and was never merged. On current master, GetExtentInfoFromDisk() still initializes IsDeleted: false without checking whether the extent is already in the delete tracking map, which is the issue this PR addresses.

If this bug is still reproducible, please open a focused fresh PR against current master (the change is small and had test coverage in the original PR).

Closing as stale. Happy to review a rebased version!

PR #3082, opened 2024-02-06
style(sdk): remove is timeout header from client

sejust:
remove is timeout header from client


tangdeyi (approved):
LGTM

PR #3096, opened 2024-02-19
feat(master): minor code and comment refactoring

### LINKED ISSUE #3097: [Enhancement]: Code modifications for readability and code quality
### Contact Details

yj.yoo@okestro.com

### What would you like to be added?

I am interested in this project and while studying, I think minor code modifications are necessary.

### Why is this needed?

I think minor code and comment modifications are necessary for readability and code quality control.

### Anything else?

_No response_

yj-yoo:

- remove redundant 'return' state
- simplify conditional expressions
- remove ineffectual assignment
  - 'fdstore.go, line 280' It will not be referenced again in the 'for' loop it belongs to, and will then be overwritten with a new value in the 'os.OpenFile()' function.
  - 'fdstore.go, line 303' Handling of the 'fmt.Errorf()' function ends without processing
  - 'io.gok line 114' slice is initialized automatically, and since it is not a pointer, it does not use nil.
  - 'flag.go, line 740' This could directly return the return value of 'fmt.Errorf()' (actually it doesn't seem necessary, but I can't figure out the intent, so it preserves the meaning of the code)
  - 'serve.go, line 1840' It is continuously initialized within the for loop and will not be referenced beyond the line in use.
- fix replaceable functions
- fix return that doesn't reach
- fix typo
- edit comments

fixes #3097

**Release note**:

```release-note
```

### DIFF
--- a/fdstore/fdstore.go
+++ b/fdstore/fdstore.go
@@ -277,7 +277,6 @@ func doDump(filePathes string) {
 
 		rsize, err = nodeListFile.Read(data)
 		if rsize == 0 || err == io.EOF {
-			err = nil
 			break
 		}
 
@@ -300,7 +299,7 @@ func doDump(filePathes string) {
 
 	handleListFile, err := os.OpenFile(pathes[1], os.O_RDONLY, 0o644)
 	if err != nil {
-		err = fmt.Errorf("failed to open handles list file: %v\n", err)
+		fmt.Fprintf(os.Stderr, "failed to open handles list file: %v\n", err)
 		return
 	}
 	defer handleListFile.Close()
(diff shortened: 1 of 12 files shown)


fdstore/fdstore.go
```diff
@@ -300,7 +299,7 @@ func doDump(filePathes string) {
 
 	handleListFile, err := os.OpenFile(pathes[1], os.O_RDONLY, 0o644)
 	if err != nil {
-		err = fmt.Errorf("failed to open handles list file: %v\n", err)
+		fmt.Errorf("failed to open handles list file: %v\n", err)
```
sejust, line 302: fmt.Fprintf ??

sejust (changes requested):

fdstore/fdstore.go
yj-yoo: ok. I applied it. thanks

sejust (approved):

leonrayang (approved):
LGTM

PR #3111, opened 2024-02-21
perf(master): move preload dp capacity calculation to the background and execute it periodically.

true1064:
…impleView(), to reduce CPU overload

perf(master): move preload dp capacity calculation to the background and execute it periodically.


leonrayang (approved):
LGTM

Victor1319 (approved):
LGTM

PR #3134, opened 2024-02-26
fix(master): If a volume is deleted, the corresponding dp should be s…

bboyCH4:
…et to offline successfully.
fix(master): If a volume is deleted, the corresponding dp should be set to offline successfully.


Victor1319 (approved):
LGTM

true1064 (approved):

PR #3137, opened 2024-02-27
feat(patch):  Apply the patches added after the release of 3.3.1 to 3.3.2

leonrayang:

feat(patch):  Apply the patches added after the release of 3.3.1 to 3.3.2

fix(fsck): return error when get connect failed.
feat(datanode): support delete dp replica even if not exist when req force is true.
enhance(master): not calculate preload dp capacity every time in newSimpleView(), to reduce CPU usage
feat(fsck): support calcuate bad extent concurrently.
……


Victor1319 (approved):
LGTM

PR #3149, opened 2024-02-28
fix(data): when traversing the replicas slice, avoid panics caused by… changes in the slice length

true1064:

fix(data): when traversing the replicas slice, avoid panics caused by… changes in the slice length


leonrayang (approved):
LGTM

bboyCH4 (approved):

PR #3174, opened 2024-03-05
fix(master): Modify the execution order of deletion after freezing to avoid possible early deletion pitfalls.

shuqiang-zheng:
Modify the execution order of deletion after freezing to avoid possible early deletion pitfalls.

fixes #

**Release note**:

```release-note
```


true1064 (approved):

leonrayang (approved):
LGTM

PR #3199, opened 2024-03-07
fix(lcnode): fix lcnode panic and reduce task allocation time

honeyvinnie:

fix lcnode panic and reduce task allocation time


tangdeyi (approved):

yhjiango (approved):
LGTM

Issue #3226, opened 2024-03-11
[Bug]: Client doesn't handle try again response correctly


NaturalSelect:
### Contact Details

huangzhibin1@oppo.com

### Is there an existing issue for this?

- [X] I have searched all the existing issues

### Priority

low (Default)

### Environment

```markdown
- CubeFS version: 3.3.1
- Deployment mode(docker or standalone or cluster):
- Dependent components:
- OS kernel version(Ubuntu or CentOS):
- CPU/Memory:
- Others:
```

### Current Behavior

In the client, the try again error code is treated as a normal error, which may quickly exhaust the number of retries under heavy load.

PR #3227, opened 2024-03-11
fix(data): rewrite disk select algorithm and client re-try handler

### LINKED ISSUE #3225: [Bug]: Many small data partitions will be placed on the same disk, causing the disk to be overloaded with subsequent writes
### Contact Details

huangzhibin1@oppo.com

### Priority

low (Default)

### Environment

```markdown
- CubeFS version: 3.3.1
- Deployment mode(docker or standalone or cluster):
- Dependent components:
- OS kernel version(Ubuntu or CentOS):
- CPU/Memory:
- Others:
```

### Current Behavior

The current algorithm only considers the current space utilization of the disk:
```go
func (manager *SpaceManager) minPartitionCnt(decommissionedDisks []string) (d *Disk) {
	manager.diskMutex.Lock()
	defer manager.diskMutex.Unlock()
	var (
		minWeight     float64
		minWeightDisk *Disk
	)
	decommissionedDiskMap := make(map[string]struct{})
	for _, disk := range decommissionedDisks {
		decommissionedDiskMap[disk] = struct{}{}
	}
	minWeight = math.MaxFloat64
	for _, disk := range manager.disks {
		if _, ok := decommissionedDiskMap[disk.Path]; ok {
			log.LogInfof("action[minPartitionCnt] exclude decommissioned disk[%v]", disk.Path)
			continue
		}
		if disk.Status != proto.ReadWrite {
			continue
		}
		diskWeight := disk.getSelectWeight()
		if diskWeight < minWeight {
			minWeight = diskWeight
			minWeightDisk = disk
		}
	}
	if minWeightDisk == nil {
		return
	}
	if minWeightDisk.Status != proto.ReadWrite {
		return
	}
	d = minWeightDisk
	return d
}

func (d *Disk) getSelectWeight() float64 {
	return float64(atomic.LoadUint64(&d.Allocated)) / float64(d.Total)
}
```

When there are too many small data partitions, the disk is easily overloaded with subsequent writes.

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### CubeFS Log

_No response_

### Anything else? (Additional Context)

_No response_

NaturalSelect:

fixes #3225 #3226 #3231

**Release note**:

```release-note
```

### DIFF
--- a/datanode/disk.go
+++ b/datanode/disk.go
@@ -238,6 +238,7 @@ func (d *Disk) CanWrite() bool {
 		return true
 	}
 
+	log.LogErrorf("[CanWrite] disk(%v) not writable, status(%v) reject write(%v) total space(%v) disk rdonly space(%v) used(%v) reserved space(%v)", d.Path, d.Status, d.RejectWrite, d.Total, d.DiskRdonlySpace, d.Used, d.ReservedSpace)
 	return false
 }
 
@@ -708,10 +709,6 @@ func (d *Disk) updateDisk(allocSize uint64) {
 		return
 	}
 	d.Available = d.Available - allocSize
-}
-
-func (d *Disk) getSelectWeight() float64 {
-	return float64(atomic.LoadUint64(&d.Allocated)) / float64(d.Total)
 }
 
 func (d *Disk) AddDiskErrPartition(dpId uint64) {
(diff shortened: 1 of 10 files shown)


NaturalSelect:
split to multi-prs

PR #3233, opened 2024-03-11
fix(meta): create dentry concurrency cause orphan inodes

sjp00556:
### Motivation

when concurrency create dentries with same parent and name will cause orphan inodes.

### Modifications

when create dentry with same parent and name: 
- if inodes are same, it will return status OpOk to client, we can treat it as successful  (fixed here: https://github.com/cubefs/cubefs/commit/f55846af3c2522b4773e5e2dd2c2871b88f80267)
- if inodes are not same, it will return proto.OpExistErr

so when we get OpExistErr, it means dentry must create failed, so we should unlink this inode to avoid orphan inode

### Types of changes

- [ ] So on...

### Verifying this change

- [x] Make sure that the change passes the testing checks.

- [ ] Master
- [ ] MetaNode
- [ ] DataNode
- [ ] ObjectNode
- [ ] AuthNode
- [ ] LcNode
- [ ] Blobstore
- [ ] Client
- [ ] Cli
- [x] SDK
- [ ] Other Tools
- [ ] Common Packages
- [ ] Dependencies

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Review Expection

- [ ] `in-two-days`
- [x] `weeks`
- [ ] `free-time`
- [ ] `whenever`

### DIFF
--- a/sdk/meta/api.go
+++ b/sdk/meta/api.go
@@ -333,10 +333,8 @@ create_dentry:
 		}
 		return nil, statusToErrno(status)
 	} else if status != statusOK {
-		if status != statusExist {
-			mw.iunlink(mp, info.Inode, mw.Client.GetLatestVer(), 0, fullPath)
-			mw.ievict(mp, info.Inode, fullPath)
-		}
+		mw.iunlink(mp, info.Inode, mw.Client.GetLatestVer(), 0, fullPath)
+		mw.ievict(mp, info.Inode, fullPath)
 		return nil, statusToErrno(status)
 	}
 	if mw.EnableSummary {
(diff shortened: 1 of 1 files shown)


sdk/meta/api.go
```diff
@@ -333,10 +333,8 @@ create_dentry:
 		}
 		return nil, statusToErrno(status)
 	} else if status != statusOK {
-		if status != statusExist {
-			mw.iunlink(mp, info.Inode, mw.Client.GetLatestVer(), 0, fullPath)
-			mw.ievict(mp, info.Inode, fullPath)
-		}
+		mw.iunlink(mp, info.Inode, mw.Client.GetLatestVer(), 0, fullPath)
```
leonrayang, line 336: Don't we need to judge this status code? If the dentry already exists when it is being created, the inode needs to be deleted because the file already exists. However, in other cases, it might be due to a network timeout. At this time, if the inode is deleted, the dentry might still remain.

leonrayang:
LGTM

leonrayang (approved):
LGTM

aaronwu2010 (commented):
length := len(rwPartitions) = 0
Then current inode is removed by mistake ?

Issue #3416, opened 2024-06-19
[Enhancement]:  `meta.NewMetaWrapper` should initialize the `epoch` with a random number


yuangcao:
### Contact Details

cyadxx@mail.ustc.edu.cn

### Is there an existing issue for this?

- [X] I have searched all the existing issues

### What would you like to be added?

I would like to suggest a feature in which the `meta.NewMetaWrapper` initializes the `epoch` with a random number. Like this: `mw.epoch = uint64(rand.Intn(len(mw.rwPartitions) + 1))`.

### Why is this needed?

I use the [mdtest with CubeFS backend support](https://github.com/yuangcao/ior/blob/main/src/aiori-CUBEFS.c) to measure the throughput of CubeFS meta subsystem. Each mdtest process will create its own CubeFS client with the `MetaWrapper` having the same `epoch`.

Each mdtest process creates the same directory tree (one dir contains millions of files). In this situation, all mdtest processes will create their root dir inode in the same meta partition, because all the mdtest processes' CubeFS client have the same `epoch`. Then all the files' dentry will be created in the same meta partition, which will cause significant performance degradation.

PR #3417, opened 2024-06-19
fix(sdk): NewMetaWrapper initializes the epoch with a random number.

yuangcao:
close: #3416

Fixes: #3416

### Motivation

I use the [mdtest with CubeFS backend support](https://github.com/yuangcao/ior/blob/main/src/aiori-CUBEFS.c) to measure the throughput of CubeFS meta subsystem. Each mdtest process will create its own CubeFS client with the MetaWrapper having the same epoch.

Each mdtest process creates the same directory tree (one dir contains millions of files). In this situation, all mdtest processes will create their root dir inode in the same meta partition, because all the mdtest processes' CubeFS client have the same epoch. Then all the files' dentry will be created in the same meta partition, which will cause significant performance degradation.

### Modifications

``` text
In the NewMetaWrapper function, I assign a random value to the mw.epoch. The value is in the range of [0, len(mw.rwPartitions)].
```

### Types of changes

- [ ] So on...

### Verifying this change

*(or)*

*(or)*

*(example:)*

*(or)*

- [ ] Master
- [ ] MetaNode
- [ ] DataNode
- [ ] ObjectNode
- [ ] AuthNode
- [ ] LcNode
- [ ] Blobstore
- [ ] Client
- [ ] Cli
- [x] SDK
- [ ] Other Tools
- [ ] Common Packages
- [ ] Dependencies

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [x] `doc-not-needed` 
- [ ] `doc-complete` 

### Review Expection

- [ ] `in-two-days`
- [ ] `weekly`
- [ ] `free-time`
- [x] `whenever`

### DIFF
--- a/sdk/meta/meta.go
+++ b/sdk/meta/meta.go
@@ -16,6 +16,7 @@ package meta
 
 import (
 	gerrors "errors"
+	"math/rand"
 	"sync"
 	"syscall"
 	"time"
@@ -240,6 +241,8 @@ func NewMetaWrapper(config *MetaConfig) (*MetaWrapper, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	mw.epoch = uint64(rand.Intn(len(mw.rwPartitions) + 1)) // Choose the init mp randomly.
 
 	go mw.updateQuotaInfoTick()
 	go mw.refresh()


leonrayang (approved):
LGTM

sejust (approved):

PR #3614, opened 2024-12-26
fix(raft): fix apply not return pool.

z-bb:
Fixes: raft

### Motivation
when using readIndex, it was found that 'apply' is not return poll and unexpected situations may occur in concurrent situations


leonrayang (approved):
LGTM

Victor1319 (approved):
LGTM

PR #3619, opened 2025-01-02
fix(client): only delete src dir directly if dstPath exists in trash and fix error log in renameToTrashTempFile

M1eyu2018:
fix(client): only delete src dir directly if dstPath exists in trash and fix error log in renameToTrashTempFile

### Motivation

When the trash policy is enabled and I delete the same directory twice, there are some error logs which is 'no such file or directory'. Because the same directory is in trash after the first deletion, so it will be deleted directly for the second time. However, after this direct deletion, it will be renamed to trash again which is unreasonable and causes 'no such file or directory' during rename.

### Modifications

When the same directory is in trash after the first deletion, just only delete the directory directly, don't need to rename to trash again.

### Types of changes

- [ ] So on...

- [ ] Master
- [ ] MetaNode
- [ ] DataNode
- [ ] ObjectNode
- [ ] AuthNode
- [ ] LcNode
- [ ] Blobstore
- [x] Client
- [ ] Cli
- [ ] SDK
- [ ] Other Tools
- [ ] Common Packages
- [ ] Dependencies

### Review Expection

- [x] `in-two-days`
- [ ] `weekly`
- [ ] `free-time`
- [ ] `whenever`

### DIFF
--- a/sdk/meta/trash.go
+++ b/sdk/meta/trash.go
@@ -228,7 +228,7 @@ func (trash *Trash) MoveToTrash(parentPathAbsolute string, parentIno uint64, fil
 				if err != nil {
 					return err
 				}
-				break
+				return nil
 			}
 		} else {
 			log.LogDebugf("action[MoveToTrash]break")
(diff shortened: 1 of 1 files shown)


leonrayang:
Thanks @M1eyu2018 for this trash policy fix!

We're closing this PR because it has been open since early 2025 targeting release-3.4.0 and was never merged as a PR. However, the core fix — when a directory with the same name already exists in trash, delete the source directly instead of attempting another rename — is already present on current master in sdk/meta/trash.go (MoveToTrash + deleteSrcDirDirectly).

Closing as superseded. Thank you!

PR #3622, opened 2025-01-03
fix(raft): fix readIndex not set index.

z-bb:
Fixes: 

Main Issue: 

### Motivation

When readIndex is supported, if committedEntries is equal to 0 and readIndexs is not 0, an apply will be obtained from the pool, but the index is not set, which may cause the index to be greater than curApplied, resulting in data loss.

### Modifications

``` text
Added readindex applied
```

### Types of changes

- [ ] So on...

### Verifying this change

*(or)*

*(or)*

*(example:)*

*(or)*

### DIFF
--- a/depends/tiglabs/raft/raft.go
+++ b/depends/tiglabs/raft/raft.go
@@ -648,6 +648,7 @@ func (s *raft) apply() {
 		}
 		apply := pool.getApply()
 		apply.readIndexes = readIndexes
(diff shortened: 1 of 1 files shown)


leonrayang (approved):
LGTM

leonrayang:
@z-bb Could you please explain the use cases for readindex?

PR #3748, opened 2025-03-27
fix(client_kernel) When the client is compatible with parsing the server message, the time zone is set to UTC.

zcola:
1. The log clearly analyzes the failed value of mt.
2. When the client is compatible with parsing the server message, the time zone is set to UTC.


aaronwu2010 (approved):
LGTM

PR #3749, opened 2025-03-27
fix(cli): migrate metanode cmd cannot work with default limit mp count

liubingxing:
Fixes: #xyz

Main Issue: #xyz

### Motivation

migrate metanode with cli cmd can not work
image

changed by this PR
[enhance(master): mp migrate need limit the parrallel count](https://github.com/cubefs/cubefs/pull/2752/commits/c137a273d9595ef200c65ef246860f33e25f24b6)

### Modifications

``` text
I think we can do the following modifications:
1、migrate metanode cli set default mp count with 0 but not maxMpCount
2、remove the check for limit value as it is specified by the user 
      and the cli command will detect that this value does not exceed 15

```

### Types of changes

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] So on...

### Verifying this change

*(or)*

*(or)*

*(example:)*

*(or)*

- [ ] Master
- [ ] MetaNode
- [ ] DataNode
- [ ] ObjectNode
- [ ] AuthNode
- [ ] LcNode
- [ ] Blobstore
- [ ] Client
- [ ] Cli
- [ ] SDK
- [ ] Other Tools
- [ ] Common Packages
- [ ] Dependencies

### Documentation

- [ ] `doc` 
- [ ] `doc-required` 
- [ ] `doc-not-needed` 
- [ ] `doc-complete` 

### Review Expection

- [ ] `in-two-days`
- [ ] `weekly`
- [ ] `free-time`
- [ ] `whenever`

### Matching PR in forked repository

PR in forked repository:

### DIFF
--- a/cli/cmd/metanode.go
+++ b/cli/cmd/metanode.go
@@ -26,7 +26,7 @@ import (
 const (
 	cmdMetaNodeUse   = "metanode [COMMAND]"
 	cmdMetaNodeShort = "Manage meta nodes"
-	mpMigrateMax     = 15
+	mpMigrateMax     = 3
 )
 
 func newMetaNodeCmd(client *master.MasterClient) *cobra.Command {
@@ -173,7 +173,7 @@ func newMetaNodeMigrateCmd(client *master.MasterClient) *cobra.Command {
 			src = args[0]
 			dst = args[1]
 			if optCount > mpMigrateMax || optCount <= 0 {
-				stdout("Migrate mp count should between [1-15]\n")
+				stdout("Migrate mp count should between [1-3]\n")
 				return
 			}
 			if err = client.NodeAPI().MetaNodeMigrate(src, dst, optCount, clientIDKey); err != nil {
--- a/cli/cmd/metanode.go
+++ b/cli/cmd/metanode.go
@@ -20,13 +20,13 @@ import (
 
 	"github.com/cubefs/cubefs/proto"
 	"github.com/cubefs/cubefs/sdk/master"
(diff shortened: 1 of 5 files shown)


cli/cmd/metanode.go
```diff
@@ -172,10 +171,6 @@ func newMetaNodeMigrateCmd(client *master.MasterClient) *cobra.Command {
 			}()
 			src = args[0]
 			dst = args[1]
-			if optCount > mpMigrateMax || optCount <= 0 {
-				stdout("Migrate mp count should between [1-15]\n")
```
leonrayang, line 176: Both here and on the server side should retain it, and use the variable defaultMigrateMpCnt. This variable can be moved to the util module.

master/api_service.go
```diff
@@ -5157,12 +5157,6 @@ func (m *Server) migrateMetaNodeHandler(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
-	if limit > defaultMigrateMpCnt {
```
leonrayang, line 5160: There should be a limit on the server side. According to experience, a concurrency of 3 is already quite a lot.

liubingxing: OK, I will fix it.

cli/cmd/metanode.go
liubingxing: fix it

leonrayang (approved):
LGTM

Victor1319 (approved):
LGTM

PR #4067, opened 2026-07-22
fix(objectnode): use full ISO8601 timestamp in chunked SigV4 chunk signing

gschuurman:
### Motivation

`buildSignatureChunk` in `objectnode/auth_header.go` passed the short credential-scope date (`cred.Date`, e.g. `"20260722"`) to `NewSignChunkedReader` as each chunk's signing datetime, instead of the full ISO8601 request timestamp (`cred.TimeStamp`, e.g. `"20260722T085745Z"`) that AWS's chunked SigV4 spec requires there (per [sigv4-streaming.html](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html), and confirmed against `aws-sdk-java-v2`'s own `SigV4ChunkExtensionProvider#getStringToSign`, which uses `credentialScope.getDatetime()` -- the full timestamp).

This made every per-chunk signature verification fail unconditionally for any client using chunked/streaming SigV4 uploads, independent of client SDK version -- reproduced with Hadoop's S3A connector across `hadoop-aws` 3.4.2/3.4.3 and AWS SDK v2 bundle 2.29.52/2.35.4, and confirmed fixed against the same client stack after this change.

### Modifications

One-line fix in `buildSignatureChunk`: pass `cred.TimeStamp` (full ISO8601 timestamp) instead of `cred.Date` (short date) into `NewSignChunkedReader`, matching the datetime AWS's spec and every SDK actually use for the per-chunk string-to-sign.

Adds `TestHeaderAuthV4ChunkedBodySignatureMatch`, a regression test that exercises the actual buggy code path (`HeaderAuth.SignatureMatch` plus reading the wrapped chunked body through), using the same official AWS chunked-upload example (`examplebucket/chunkObject.txt`) already used by `TestHeaderAuthV4SignatureMatch`'s "Chunked PUT Object" case and `TestSignChunkedReader`. The existing `TestSignChunkedReader` alone could never have caught this bug: it calls `NewSignChunkedReader` directly with the correct full-timestamp value, bypassing the buggy call site in `buildSignatureChunk` entirely. Confirmed this new test fails with the exact real-world error (`"signature of chunk does not match"`) when the bug is reintroduced, and passes with the fix, without affecting any other existing test in the package.

**Not a breaking change in practice**: every standards-compliant SDK already computes chunk signatures using the full timestamp (verified against `aws-sdk-java-v2`'s own source), so no compliant client could have ever successfully completed a chunked upload against ObjectNode before this fix -- every such upload was already failing 100% of the time with `"signature of chunk does not match"`, regardless of SDK/version. This fixes previously-broken functionality; it can't regress a client that was never able to complete a chunked upload in the first place. Only a hypothetical non-standard client deliberately reverse-engineered to sign chunks the same (incorrect) way ObjectNode previously expected would be affected, which no known SDK does. Clients that avoid chunked encoding (e.g. older SDKs defaulting to a single non-chunked signed PUT for small objects) never exercised this code path and are unaffected either way.

### Types of changes

- [ ] So on...

### Verifying this change

- `TestHeaderAuthV4ChunkedBodySignatureMatch` (new) exercises the real bug: it builds the same official AWS chunked-upload request `TestHeaderAuthV4SignatureMatch` already uses, calls `SignatureMatch` (which wires up the chunked body reader), then actually reads the decoded body through and asserts it succeeds and matches the expected content. Confirmed this test fails with `"signature of chunk does not match"` when the one-line fix is reverted, and passes with it applied -- with `go test ./objectnode/... -run 'TestHeaderAuthV4ChunkedBodySignatureMatch|TestHeaderAuthV4SignatureMatch|TestSignChunkedReader|TestParseSignChunkedHeader'` green in both directions as expected.
- Also verified end-to-end against a real running CubeFS cluster: a real Spark write (Hadoop `hadoop-aws` 3.4.2 and 3.4.3, AWS SDK v2 bundle 2.29.52 and 2.35.4) through `ObjectNode`'s S3 API failed with `AWSStatus500Exception: ... signature of chunk does not match` before this fix, and succeeded (write + readback verified) after rebuilding `ObjectNode` with it applied -- both against an isolated standalone instance and the real multi-replica deployment.

- [ ] Master
- [ ] MetaNode
- [ ] DataNode
- [x] ObjectNode
- [ ] AuthNode
- [ ] LcNode
- [ ] Blobstore
- [ ] Client
- [ ] Cli
- [ ] SDK
- [ ] Other Tools
- [ ] Common Packages
- [ ] Dependencies

### Documentation

- [ ] `doc`
- [ ] `doc-required`
- [x] `doc-not-needed`
- [ ] `doc-complete`

No user-facing behavior or configuration changes -- this only makes ObjectNode's chunk-signature verification match the AWS spec it was already supposed to implement.

### DIFF
--- a/objectnode/auth_header.go
+++ b/objectnode/auth_header.go
@@ -233,7 +233,15 @@ func (auth *HeaderAuth) buildSignatureChunk(secretKey string) string {
 
 	signature := calculateSignature(signingKey, auth.stringToSign)
 
-	auth.request.Body = NewSignChunkedReader(auth.request.Body, signingKey, scope, cred.Date, signature)
+	// Per-chunk signing (https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-streaming.html)
+	// uses the full ISO8601 request timestamp ("20060102T150405Z") in each chunk's
+	// string-to-sign, not the short credential-scope date ("20060102") -- every AWS SDK
+	// builds it this way (confirmed directly against aws-sdk-java-v2's
+	// SigV4ChunkExtensionProvider#getStringToSign, which uses credentialScope.getDatetime(),
+	// the full timestamp). Passing cred.Date here instead of cred.TimeStamp made every
+	// chunk signature verification fail unconditionally for any client that actually uses
+	// chunked/streaming SigV4 uploads, independent of SDK version.
+	auth.request.Body = NewSignChunkedReader(auth.request.Body, signingKey, scope, cred.TimeStamp, signature)
 
 	return signature
 }
--- a/objectnode/auth_header_test.go
+++ b/objectnode/auth_header_test.go
@@ -15,8 +15,11 @@
 package objectnode
 
 import (
+	"bytes"
 	"fmt"
+	"io"
 	"net/http/httptest"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


gschuurman:
@jzeng4 @Cloudstriff  Do you have the opportunity to check this PR? This will solve compatibility issues with the latest S3 SDK's

tangdeyi (dismissed):

Cloudstriff (dismissed):

tangdeyi:
The commit message is formatted incorrectly; please correct it. @gschuurman 
image

gschuurman:
> The commit message is formatted incorrectly; please correct it. @gschuurman image

Thanks, I fixed the Sign off

tangdeyi:
> > The commit message is formatted incorrectly; please correct it. @gschuurman image
> 
> Thanks, I fixed the Sign off

This problem still exists. @gschuurman  https://github.com/cubefs/cubefs/pull/4067/checks?check_run_id=93766598990
Summary
Commit sha: [fed91c4](https://github.com/cubefs/cubefs/pull/4067/commits/fed91c41cc1e7f0aea0d088c4d75d4b3c01840d7), Author: Glenn Schuurman, Committer: tangdeyi; Expected "Glenn Schuurman [gschuurman@scamander.com](mailto:gschuurman@scamander.com)", but got "Glenn Schuurman [glennschuurman@gmail.com](mailto:glennschuurman@gmail.com)".

gschuurman:
@tangdeyi It should now finally be fixed, I had the wrong email configured for this repo.

tangdeyi (approved):

honeyvinnie (approved):

Victor1319 (approved):
LGTM

PR #4093, opened 2026-08-28
fix(proto): widen the version list buffer size so a peer count above 3855 is readable

tzh476:
## What

`ReadFromConnWithVer` reads a `uint16` count off the connection and sizes the version-list buffer with it:

```go
cnt := binary.BigEndian.Uint16(cntByte)
...
verData := make([]byte, cnt*verInfoCnt)
if _, err = io.ReadFull(c, verData); err != nil { ... }
err = p.UnmarshalVersionSlice(int(cnt), verData)
```

`cnt` is a `uint16` and `verInfoCnt` is 17, so `cnt*verInfoCnt` is evaluated in `uint16` arithmetic and **wraps for any `cnt` above 3855**. `io.ReadFull` is then given a buffer far shorter than the packet body, the read comes up short, and the packet is rejected.

This is not an over-allocation — it is a truncation, and the practical effect is that **a version list longer than 3855 entries cannot be received at all**. The count comes from a peer, so it is reachable from the network.

## Measured in the `proto` package

```
cnt=3855   cnt*verInfoCnt (uint16) = 65535   needed 65535     ok
cnt=3856   cnt*verInfoCnt (uint16) = 16      needed 65552
cnt=30000  cnt*verInfoCnt (uint16) = 51248   needed 510000
cnt=65535  cnt*verInfoCnt (uint16) = 65519   needed 1114095
```

## Scope of the impact

`UnmarshalVersionSlice` reads through a `bytes.Buffer` with `binary.Read`, so a short
buffer makes it return `EOF` rather than read out of bounds. I checked this directly
rather than assuming it:

```
UnmarshalVersionSlice(3856,  make([]byte, 16))     -> EOF, 0 items
UnmarshalVersionSlice(65535, make([]byte, 65519))  -> unexpected EOF, 0 items
```

So this is a **correctness and availability** bug, not a memory-safety one. An earlier
revision of this PR described it as reading past the end of the buffer; that was wrong
and I have corrected it here and in the code comment.

## Fix

Widen the multiplication:

```go
verData := make([]byte, int(cnt)*verInfoCnt)
```

`io.ReadFull` still bounds how much is actually read from the connection, so this does not let a peer make the process read more than it sends — it only makes the buffer size agree with the count that was announced.

## Testing

`proto/packet_verlist_test.go`:

- **`TestReadFromConnWithVerCount`** drives the real `ReadFromConnWithVer` over a
  `net.Pipe`, writing a header with `VersionListFlag` set followed by the `uint16`
  count and `cnt` entries, for `cnt` of 0 / 1 / 100 / **3855 / 3856** / 5000. It asserts
  the packet parses and that `VerList` holds exactly the entries that were sent.
- **`TestVersionListBufferSizeDoesNotWrap`** states the arithmetic directly, so the
  boundary stays documented if the wire format changes.

Control run — with `proto/packet.go` reverted to the unwidened multiplication and these
tests kept:

```
--- FAIL: TestReadFromConnWithVerCount/#04   cnt=3856 must parse
--- FAIL: TestReadFromConnWithVerCount/#05   cnt=5000 must parse
```

`cnt <= 3855` passes on both trees, so the failure lands exactly on the overflow
boundary. With the fix applied, `go test ./proto/` passes in full; `gofmt` and
`go vet ./proto/` are clean.

## Disclosure

This change was prepared with AI assistance. Every number above is from a run on the
diff as submitted, and I can explain and revise it during review.

### DIFF
--- a/proto/packet.go
+++ b/proto/packet.go
@@ -1216,7 +1216,12 @@ func (p *Packet) ReadFromConnWithVer(c net.Conn, timeoutSec int) (err error) {
 		cnt := binary.BigEndian.Uint16(cntByte)
 		log.LogDebugf("action[ReadFromConnWithVer] op %s verseq %v, extType %d, cnt %d",
 			p.GetOpMsg(), p.VerSeq, p.ExtentType, cnt)
-		verData := make([]byte, cnt*verInfoCnt)
+		// cnt is a uint16 read from the connection, and verInfoCnt is 17, so
+		// cnt*verInfoCnt is computed in uint16 arithmetic and wraps: a cnt of
+		// 3856 asks for 65552 bytes but yields a 16 byte buffer, and the short
+		// read below then rejects the packet. Widen the multiplication so the
+		// buffer matches what cnt actually describes.
+		verData := make([]byte, int(cnt)*verInfoCnt)
 		if _, err = io.ReadFull(c, verData); err != nil {
 			err = fmt.Errorf("reqId(%v) opCode(%v) protoVer(%v) remote(%v), read ver slice from conn failed: %v",
 				p.GetReqID(), p.GetOpMsg(), p.ProtoVersion, c.RemoteAddr(), err)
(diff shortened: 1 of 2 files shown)


Cloudstriff (approved):
LGTM

Victor1319 (approved):
LGTM

